PR #3, opened 2020-03-23
Fix buffer size for .toString()

beegee-tokyo:
Stupid mistake if you define a buffer with the length of 6 but want to snprintf 8 or more digits to it.

See https://github.com/espressif/arduino-esp32/pull/3836


h2zero:
At least it was snprintf :), though sometimes i'd rather it crash.

PR #6, opened 2020-03-31
fix typos

tobozo:


h2zero:
Wow those typos never go away once you make the first one haha. Thanks for this!

PR #14, opened 2020-04-20
Adding getDataLength()

tobozo:


h2zero:
I think we should define the function in the cpp file. I know it’s not necessary but it keeps the convention the rest of the library follows.

tobozo:
for some reason my text editor wants to insert that newline at the end of the files, sorry about that :-)

PR #26, opened 2020-04-30
Stop copying read values in characteristic

Jeroen88:
readRawData() returns a pointer to m_value now, instead of keeping a second copy of the same data. Small API change, if no data is available, an empty string is returned instead of a NULL. Also readValue() returns a reference to m_value now, thus saving a copy constructor and a second copy of the same string in the sketch.

I am curious what you think of this improvement. There might be more locations in the library to do similar things.

In general though, I believe it is a bad habit to store temporary (read) values in the class instance. This PR is not ideal but an improvement of the current situation.


h2zero:
I like these changes, but I think they would be better applied to the `client-long-read-write` branch as I am working on changes to the implementation of this code there, and a few of these have already been done there as well.

Jeroen88:
Okay,closing for now. I am looking forwardto the new release :)

PR #31, opened 2020-05-03
NimBLEUUID: Add operator == and != + correct equals() to test if values are set.

Jeroen88:
With these operators you now can do:

```
BLEUUID first = BLEUUID(0xebe0ccb0, 0x7a0a, 0x4b0c, 0x8a1a6ff2997da3a6);
BLEUUID second = first;
if(first == second) ....
if(first != BLEUID("180f")) ...
std::string stringValue = std::string(first);
```

Also, ::equals() did not test if the values were set. So two uninitialized values where likely not to be equal, while in my opinion they are. If one or both values to compare were unset, then initialized memory was compared. Although not harmful, this is incorrect.

This may be inspiration to add these operators to a lot more classes defined in this library.

### DIFF
--- a/src/NimBLEUUID.cpp
+++ b/src/NimBLEUUID.cpp
@@ -198,12 +198,12 @@ uint8_t NimBLEUUID::bitSize() {
  * @return True if the UUIDs are equal and false otherwise.
  */
 bool NimBLEUUID::equals(NimBLEUUID uuid) {
-    if(m_valueSet && rhs.m_valueSet) {
+    if(m_valueSet && uuid.m_valueSet) {
         if(ble_uuid_cmp(&m_uuid.u, &uuid.getNative()->u) == 0){
             return true;
         }
     }
-    return m_valueSet == rhs.m_valueSet;
+    return m_valueSet == uuid.m_valueSet;
 }
 
 
--- a/src/NimBLEUUID.cpp
+++ b/src/NimBLEUUID.cpp
@@ -229,47 +229,21 @@ ble_uuid_any_t* NimBLEUUID::getNative() {
  * A UUID can be internally represented as 16bit, 32bit or the full 128bit.  This method
  * will convert 16 or 32 bit representations to the full 128bit.
  */
-NimBLEUUID NimBLEUUID::to128() {
+NimBLEUUID &NimBLEUUID::to128() {
     // If we either don't have a value or are already a 128 bit UUID, nothing further to do.
     if (!m_valueSet || m_uuid.u.type == BLE_UUID_TYPE_128) {
         return *this;
     }
 
-    // If we are 16 bit or 32 bit, then set the 4 bytes of the variable part of the UUID.
+    // If we are 16 bit or 32 bit, then set the other bytes of the UUID.
     if (m_uuid.u.type == BLE_UUID_TYPE_16) {
-        uint16_t temp = m_uuid.u16.value;
-        m_uuid.u128.value[15] = 0;
-        m_uuid.u128.value[14] = 0;
-        m_uuid.u128.value[13] = (temp >> 8) & 0xff;
-        m_uuid.u128.value[12] = temp & 0xff;
-
+        *this = NimBLEUUID(m_uuid.u16.value, 0x0000, 0x1000, 0x800000805f9b34fb);
     }
     else if (m_uuid.u.type == BLE_UUID_TYPE_32) {
-        uint32_t temp = m_uuid.u32.value;
-        m_uuid.u128.value[15] = (temp >> 24) & 0xff;
-        m_uuid.u128.value[14] = (temp >> 16) & 0xff;
-        m_uuid.u128.value[13] = (temp >> 8) & 0xff;
-        m_uuid.u128.value[12] = temp & 0xff;
+        *this = NimBLEUUID(m_uuid.u32.value, 0x0000, 0x1000, 0x800000805f9b34fb);
     }
 
-    // Set the fixed parts of the UUID.
-    m_uuid.u128.value[11] = 0x00;
-    m_uuid.u128.value[10] = 0x00;
-
-    m_uuid.u128.value[9]  = 0x10;
-    m_uuid.u128.value[8]  = 0x00;
-
-    m_uuid.u128.value[7]  = 0x80;
-    m_uuid.u128.value[6]  = 0x00;
-
-    m_uuid.u128.value[5]  = 0x00;
-    m_uuid.u128.value[4]  = 0x80;
-    m_uuid.u128.value[3]  = 0x5f;
-    m_uuid.u128.value[2]  = 0x9b;
-    m_uuid.u128.value[1]  = 0x34;
-    m_uuid.u128.value[0]  = 0xfb;
-
-    m_uuid.u.type = BLE_UUID_TYPE_128;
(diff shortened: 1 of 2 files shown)


src/NimBLEUUID.cpp
```diff
@@ -318,4 +320,34 @@ std::string NimBLEUUID::toString() {
     return ble_uuid_to_str(&m_uuid.u, buf);
 } // toString
 
+
+NimBLEUUID & NimBLEUUID::operator=(const NimBLEUUID & other) {
+    m_uuid.u.type = other.m_uuid.u.type;
```
h2zero: This should be moved inside the `if(other.m_valueSet) {` block.

Jeroen88: You are right, but I think the copy constructor can be left out completely

src/NimBLEUUID.cpp
```diff
@@ -318,4 +320,34 @@ std::string NimBLEUUID::toString() {
...
+    if(other.m_valueSet) {
+        memcpy(m_uuid.u128.value, other.m_uuid.u128.value, 16);
```
h2zero: Maybe we should put a `swtich(other.m_uuid.type) {` then add memcpy for `u16`  and `u32` types as well.

Jeroen88:
Did you notice that for the std::string cast I return an empty string `std::string()` instead of a `std::string("<null>")`? I think this is more common, but leads to a difference with the origina api.

src/NimBLEUUID.cpp
Jeroen88: I had the same idea in the first place, but the difference is (at most) two extra compares to save 14 or 12 bytes of memcpy. I think this is all the same...

src/NimBLEUUID.cpp
```diff
@@ -318,4 +320,34 @@ std::string NimBLEUUID::toString() {
...
+    }
+
+    return m_valueSet == rhs.m_valueSet;
+}
+
+bool NimBLEUUID::operator !=(const NimBLEUUID & rhs) {
+    return !this->operator==(rhs);
+}
+
+NimBLEUUID::operator std::string() const {
```
h2zero, line 277: Not really sure about this one, cast operators are nice, but I don't really see a use case?

h2zero:
@Jeroen88 I left a couple comments. Other than that, looks good!

src/NimBLEUUID.cpp
Jeroen88: Well I rather use standard cpp, and would set ::toString() deprecated...

Jeroen88:
@h2zero thnx for the comments. Do you like these changes, or is it just water under the bridge?

h2zero:
@Jeroen88, sorry I’m at work still. Just had a minute on break to review. I’ll get back to you after I get home.

src/NimBLEUUID.cpp
h2zero: We don’t need to memcpy all 3 values, just the one defined by the type value. Also it is needed for `ble_uuid_cmp()` and `ble_uuid_to_str()` to work correctly.

h2zero:
>Did you notice that for the std::string cast I return an empty string std::string() instead of a std::string("<null>")? I think this is more common, but leads to a difference with the origina api.

Yes, and I would prefer an empty string for both actually but that’s what was in the original library. 

The thing about the string methods is they only serve the purpose of printing them to the console, which we still need to use `c_str()`. I could see value if we could call `print(uuid)` to simply things, but for its limited use adding the cast operator to just remove `toString()` from the print call just doesn’t seem worth adding the code.
I’m open to it though if there’s another use case I’m not seeing?

Other than that I like what you’ve got here and there are other places that could use similar operator overloads, address’ come to mind.

Jeroen88:
> The thing about the string methods is they only serve the purpose of printing them to the console, 

I did not check this, but I believe the strings are also used as keys for maps, and maybe other purposes.

> but for its limited use adding the cast operator to just remove toString() from the print call just doesn’t seem worth adding the code.

No harm either in keeping it in, it is just a few bytes, agree?

Further changes:
- calling operator == to simplify ::equals()
- calling operator std::string() to simplify ::toString()
- using std::reverse to simplify ::NimBLEUUID(uint8_t* pData, size_t size, bool msbFirst). ::memrcpy may be removed from it's class, if the same 'trick' is applied everywhere where ::memrcpy is used, thus saving a few bytes
- refactoring ::NimBLEUUID(std::string value) to use stroul() and stroull() instead of nimble by nimble hex conversion. The function presumes the UUID is well-formed, it does not do any checking (like the original). I removed the comment about the 16 byte string: the only possible use case for this is to use a std::string as a buffer for 16 binary bytes representing the 16 bytes of the UUID

Jeroen88:
Simplifying ::to128() using the 'new' constructor. Also returning a reference `NimBLEUUID &` instead of a `NimBLEUUID`, to avoid copying, although I am not a 100% sure if this is okay.

src/NimBLEUUID.cpp
```diff
@@ -42,50 +44,25 @@ static const char* LOG_TAG = "NimBLEUUID";
     m_valueSet = true;
     if (value.length() == 4) {
```
h2zero, line 45: I think this should be changed to use the NimBLE function so the byte order is correct.
Possibly this:
```
if (value.length() == 4 || value.length() == 8 ) {
    uint32_t buf = strtoul(value.c_str(), NULL, 16);
    int rc = ble_uuid_init_from_buf(m_uuid, buf, value.length());
    if(rc != 0) {
        NIMBLE_LOGE(LOG_TAG, "Invalid UUID");
        return;
    }
}
```
That would eliminate the first 2 conditionals.

Untested but should work.

src/NimBLEUUID.cpp
```diff
@@ -255,47 +229,21 @@ ble_uuid_any_t* NimBLEUUID::getNative() {
...
-
-    m_uuid.u128.value[5]  = 0x00;
-    m_uuid.u128.value[4]  = 0x80;
-    m_uuid.u128.value[3]  = 0x5f;
-    m_uuid.u128.value[2]  = 0x9b;
-    m_uuid.u128.value[1]  = 0x34;
-    m_uuid.u128.value[0]  = 0xfb;
-
-    m_uuid.u.type = BLE_UUID_TYPE_128;
+    m_uuid.u.type == BLE_UUID_TYPE_128;
```
h2zero: I think we can remove this.

h2zero:
Good work, looking better! The reference return in `to128()` I'm unsure of, depends on how it is being used I guess. I'll have to put this through some testing.

src/NimBLEUUID.cpp
h2zero: Nevermind, wasn't thinking, this isn't necessary.

src/NimBLEUUID.cpp
Jeroen88: You are right, I'll remove it

src/NimBLEUUID.cpp
h2zero: Looks good with the change to `:toString()`

h2zero:
Looks good to me, no issues using the reference in `to128()` that I could find. Thanks!

h2zero:
Changed the commit header for accuracy, hope you don't mind.

PR #33, opened 2020-05-05
Refactoring NimBLEAddress

Jeroen88:
- Add logging for malformed addresses
- Set invalid addresses to "00:00:00:00:00:00", so the main program can easily check if an address is correctly parsed (`if(NimBLEAddress("invalid address") == NimBLEAddress(0)) ...`)
- Add constructor directly from a (hex) uint64_t value. `NimBLEAddress(0xa4c1385def16)` yields the same address as `NimBLEAddress("a4:c1:38:5d:ef:16")`
- Remove `memrcpy()`, using standard cpp instead
- Add `operator ==`, `operator !=` and `operator std::string()` cast
- Use `operator ==` in `::equals()`
- Use `operator std::string()` in `::toString()` (also avoiding the `malloc` because the cast uses a local buffer that is declared on the stack)

### DIFF
--- a/src/NimBLEAddress.cpp
+++ b/src/NimBLEAddress.cpp
@@ -47,7 +47,7 @@ NimBLEAddress::NimBLEAddress(ble_addr_t address) {
  *
  * @param [in] stringAddress The hex representation of the address.
  */
-NimBLEAddress::NimBLEAddress(std::string stringAddress) {
+NimBLEAddress::NimBLEAddress(const std::string &stringAddress) {
     if (stringAddress.length() != 17) {
         memset(m_address, 0, sizeof m_address); // "00:00:00:00:00:00" represents an invalid address
         NIMBLE_LOGD(LOG_TAG, "Invalid address '%s'", stringAddress.c_str());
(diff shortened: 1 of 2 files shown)


src/NimBLEAddress.cpp
```diff
@@ -44,25 +48,39 @@ NimBLEAddress::NimBLEAddress(ble_addr_t address) {
...
 
 
 /**
- * @brief Constructor for compatibility with bluedrioid esp library.
+ * @brief Constructor for compatibility with bluedroid esp library.
  * @param [in] uint8_t[6] or esp_bd_addr_t struct containing the address.
  */
 NimBLEAddress::NimBLEAddress(uint8_t address[6]) {
-    NimBLEUtils::memrcpy(m_address, address, 6);
+    memcpy(m_address, address, sizeof m_address);
```
h2zero: Any reason to not just use `std::reverse_copy` here?

src/NimBLEAddress.cpp
```diff
@@ -44,25 +48,39 @@ NimBLEAddress::NimBLEAddress(ble_addr_t address) {
...
+    }
+    for(size_t index = 0; index < sizeof m_address; index++) {
+        m_address[index] = data[index];
+    }
 } // BLEAddress
 
 
 /**
- * @brief Constructor for compatibility with bluedrioid esp library.
+ * @brief Constructor for compatibility with bluedroid esp library.
```
h2zero, line 69: lol, good catch!

h2zero:
Minor comment, but looks good! I'll give it some testing, thanks 😄.

src/NimBLEAddress.cpp
Jeroen88: There is always just one good reason for not using existing implementations: being unaware of it :D. I'll change it here and in NimBLEUUID too

src/NimBLEAddress.cpp
Jeroen88: A real improvement ;)

PR #35, opened 2020-05-07
Pass parameters as const &

Jeroen88:
Pass each `NimBLEUUID`, `NimBLEAddress`, `std::string` and `uint8_t * parameter` as `const &` instead of by value. This leads to (a bit) smaller code and performance improvement.

@h2zero I could not stop myself ;)


h2zero:
Lol, I’ll have a look when I get some time. Thanks!

h2zero:
Thanks @Jeroen88, nice work on this, saves some cycles (and my time 😃 ) for sure.

PR #38, opened 2020-05-09
Conditionally compile error strings in NimBLEUtils

Jeroen88:
Introducing #define NIMBLEUTILS_NO_STRINGS to conditionally compile error strings saving 8,484 bytes of program memory (with of course less error information as drawback)

### DIFF
--- a/src/NimBLEUtils.cpp
+++ b/src/NimBLEUtils.cpp
@@ -81,6 +81,7 @@ const char* NimBLEUtils::returnCodeToString(int rc) {
     switch(rc) {
         case 0:
             return "SUCCESS";
+#ifndef NIMBLEUTILS_NO_STRINGS
         case BLE_HS_EAGAIN:
             return "Temporary failure; try again.";
         case BLE_HS_EALREADY:
@@ -355,9 +356,13 @@ const char* NimBLEUtils::returnCodeToString(int rc) {
             return "Pairing over the LE transport failed - Pairing Request sent over the BR/EDR transport in process.";
         case  (0x0500+BLE_SM_ERR_CROSS_TRANS ):
             return "BR/EDR Link Key generated on the BR/EDR transport cannot be used to derive and distribute keys for the LE transport.";
-            
         default:
             return "Unknown";
+#else // #endif // #ifndef NIMBLEUTILS_NO_STRINGS
+        default:
+            return "No error texts";
+#endif // #ifndef NIMBLEUTILS_NO_STRINGS
+
     }
 }   
 
@@ -369,6 +374,7 @@ const char* NimBLEUtils::returnCodeToString(int rc) {
  
 const char* NimBLEUtils::advTypeToString(uint8_t advType) {
     switch(advType) {
+#ifndef NIMBLEUTILS_NO_STRINGS
         case BLE_HCI_ADV_TYPE_ADV_IND :                     //0
             return "Undirected - Connectable / Scannable";
         case BLE_HCI_ADV_TYPE_ADV_DIRECT_IND_HD:            //1
@@ -381,6 +387,10 @@ const char* NimBLEUtils::advTypeToString(uint8_t advType) {
             return "Directed Low Duty - Connectable";
         default:
             return "Unknown flag";
+#else // #endif // #ifndef NIMBLEUTILS_NO_STRINGS
+        default:
+            return "No flag texts";
+#endif // #ifndef NIMBLEUTILS_NO_STRINGS
     }
     
 } // adFlagsToString
@@ -434,6 +444,7 @@ void NimBLEUtils::dumpGapEvent(ble_gap_event *event, void *arg){
  */
 const char* NimBLEUtils::gapEventToString(uint8_t eventType) {
     switch (eventType) {
+#ifndef NIMBLEUTILS_NO_STRINGS
         case BLE_GAP_EVENT_CONNECT :                    //0
             return "BLE_GAP_EVENT_CONNECT ";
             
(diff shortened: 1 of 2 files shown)


Staars:
I did not test this yet, but generally spoken: I like the idea.

src/NimBLEUtils.cpp
```diff
@@ -81,6 +81,7 @@ const char* NimBLEUtils::returnCodeToString(int rc) {
     switch(rc) {
         case 0:
             return "SUCCESS";
+#ifndef NIMBLEUTILS_NO_STRINGS
```
h2zero: Could we change this to `#ifdef NIMBLE_ENABLE_DEBUG_TEXT` and move it to just before `switch(rc)`, I think it would be better if the user had to explicitly enable rather than disable the messages.

src/NimBLEUtils.cpp
```diff
@@ -355,9 +356,13 @@ const char* NimBLEUtils::returnCodeToString(int rc) {
             return "Pairing over the LE transport failed - Pairing Request sent over the BR/EDR transport in process.";
         case  (0x0500+BLE_SM_ERR_CROSS_TRANS ):
             return "BR/EDR Link Key generated on the BR/EDR transport cannot be used to derive and distribute keys for the LE transport.";
-            
         default:
             return "Unknown";
+#else // #endif // #ifndef NIMBLEUTILS_NO_STRINGS
+        default:
+            return "No error texts";
```
h2zero: We could just return an empty string here.

src/NimBLEUtils.cpp
```diff
@@ -369,6 +374,7 @@ const char* NimBLEUtils::returnCodeToString(int rc) {
  
 const char* NimBLEUtils::advTypeToString(uint8_t advType) {
     switch(advType) {
+#ifndef NIMBLEUTILS_NO_STRINGS
```
h2zero: Same thing here as `returnCodeToString()`

src/NimBLEUtils.cpp
```diff
@@ -434,6 +444,7 @@ void NimBLEUtils::dumpGapEvent(ble_gap_event *event, void *arg){
  */
 const char* NimBLEUtils::gapEventToString(uint8_t eventType) {
     switch (eventType) {
+#ifndef NIMBLEUTILS_NO_STRINGS
```
h2zero: Same here 😃

h2zero:
@Jeroen88 LOL, you are crossing out my todo list quickly. This was to be done soon as well, I've just been focused on fixing up more functionality first.

Just a couple comments, looks good otherwise, thanks!

src/NimBLEUtils.cpp
Jeroen88: I made it reverse, because of backward compatibility, but it's equal for me.

src/NimBLEUtils.cpp
Jeroen88: Yes, or a suggestion that they can be enabled at compile time?

Jeroen88:
My pleasure!

src/NimBLEUtils.cpp
h2zero: I'd rather an empty string for cleaner logs, or "<N/A>" at most.

src/NimBLEUtils.cpp
h2zero: Not too concerned about backward compatibility with this, just seems more natural the debugging should be enabled explicitly rather than disabled.

Jeroen88:
Should not here `CONFIG_BT_NIMBLE_DEBUG` from `nimconfig.h` be used in stead off `NIMBLE_ENABLE_DEBUG_TEXT`?

h2zero:
I'm actually working on that now, need a better solution for use in IDF, I'll get back to you on this soon.

h2zero:
@Jeroen88 We will need to change option names to these:
```
CONFIG_NIMBLE_CPP_ENABLE_RETURN_CODE_TEXT
CONFIG_NIMBLE_CPP_ENABLE_GAP_EVENT_CODE_TEXT
CONFIG_NIMBLE_CPP_ENABLE_ADVERTISMENT_TYPE_TEXT
```
Then in `nimconfig.h` add:
```
#define CONFIG_NIMBLE_CPP_ENABLE_RETURN_CODE_TEXT 0
#define CONFIG_NIMBLE_CPP_ENABLE_GAP_EVENT_CODE_TEXT 0
#define CONFIG_NIMBLE_CPP_ENABLE_ADVERTISMENT_TYPE_TEXT 0
```
Also we will probably need to change the conditional checks to `#if` instead of `#ifdef` to test the define value.

Jeroen88:
Since you're working on it and know exactly what you want, isn't it easier if you do it?

h2zero:
I was working on it in IDF so I thought I'd post what I got here. But sure I'll do it.

Jeroen88:
Thank you! :)

h2zero:
Done, if it works for you could you close this please?

Jeroen88:
I checked:

Without return code texts
```
Sketch uses 547710 bytes (41%) of program storage space. Maximum is 1310720 bytes.
Global variables use 25648 bytes (7%) of dynamic memory, leaving 302032 bytes for local variables. Maximum is 327680 bytes.
```

With return code texts:
```
Sketch uses 555430 bytes (42%) of program storage space. Maximum is 1310720 bytes.
Global variables use 25648 bytes (7%) of dynamic memory, leaving 302032 bytes for local variables. Maximum is 327680 bytes.
```

So 7,720 bytes difference. It is working, thnx!

PR #39, opened 2020-05-09
EXPERIMENTAL: Introducing NimBLEConfig.h, adding the possibilty to compile client only, server only or both

Jeroen88:
Introducing `NimBLEConfig.h`, adding the possibilty to compile client only, server only or both:
- Compiling _client only_ saves 15,440 bytes of program memory and 176 bytes of global variables
- Compiling _server only_ saves 27,364 bytes of program memory and 144 bytes of global variables
both compared to current master at the time of creating this PR, compiling _both client and server_.

This PR should be checked. I assigned class files to either client or server if I was pretty sure of it. But there might be more files that can be excluded. Maybe even extra `#defines` should be introduced for files like `NimBLEBeacon`, `NimBLEEddystoneTLM` and `NimBLEEddystoneURL`? Also, in `NimBLEDevice` functions and variables were sorted out, but there might be errors in here too.


Jeroen88:
@h2zero Keeping you busy :D

h2zero:
@Jeroen88 Damn... how do you read my mind? I was just thinking about this today while I was contemplating different options for upcoming things like BLE Mesh, which wouldn't require server/client code at all (big project). Thanks for beating me to it! 😄

h2zero:
@Jeroen88 I'm wondering if we should tie this with `nimconfig.h` using the defines [here](https://github.com/h2zero/NimBLE-Arduino/blob/ee9bdaf6d00bb88381802b900ca98647a467a49f/src/nimconfig.h#L10) might make an even larger difference if we can remove the NimBLE related code too.

Jeroen88:
@h2zero Lol, I'm a psychic ;)

Not a lot more 'big' changes coming, there's just one with some impact left in my mind, and a few ideas that I might propose PR's for.
Talking about changes, could you put changing the `std::map`'s into `std:vector`'s of the peripheral on your to do list? I've been using Bluetooth for just a few days now. I want to use the library as a central and have no plans in the near future to create a peripheral. So I have little experience and no test peripheral. Not a good starting point for such a change. So, could you do it? It would be nice to have the same skeleton for both remote and local objects.
I was also thinking about introducing two classes, NimBLEObject and NimBLEObjectWithChildren, the latter derived from the first, and deriving (remote)descriptor, (remote)characteristic and (remote)service from these classes. This saves a little bit of code, since all of these objects share methods for UUIDs, handles, etc, and most for child objects. But this adds an extra layer of abstracting with only small benefits I think. What is your opinion?

I saw `nimconfig.h`, but I wasn't sure if this file originated from the original BLE library. Sure I can put the `#define`'s there.
Are there already `#define`'s in `nimconfig.h` for this? Or should I create new ones? Do you have suggestions for these identifiers? I was also thinking of changing `NIMBLE_INCLUDE_SERVER` into something like `NIMBLE_INCLUDE_PERIPHERAL` and `NIMBLE_INCLUDE_CLIENT` into `NIMBLE_INCLUDE_CENTRAL`, thus better reflecting their purposes.

h2zero:
@Jeroen88, Yes I will deal with the peripheral side code changes. 

`nimconfig.h` is an Adruino specific file created since we don't have access to menuconfig here to turn NimBLE host stack on and allows us to setup NimBLE options as we would in menuconfig. 
What we can do is put in `NimBLEDevice.h` at this [line](https://github.com/h2zero/NimBLE-Arduino/blob/ee9bdaf6d00bb88381802b900ca98647a467a49f/src/NimBLEDevice.h#L19) :
```
#ifdef ARDUINO_ARCH_ESP32
#include "nimconfig.h"
#endif 
```
Then we could use `#ifdef  CONFIG_BT_NIMBLE_ROLE_CENTRAL` and `#ifdef CONFIG_BT_NIMBLE_ROLE_PERIPHERAL` to conditionally compile class code as well as stack code. This would allow it to work in IDF as well.

Jeroen88:
@h2zero done!

h2zero:
Haven't tested this yet, but i'm curious if you saw any change (further code reduction) or problems after setting `CONFIG_BT_NIMBLE_ROLE_PERIPHERAL` to 0?

Jeroen88:
I don't know what you mean exactly, but I didn't set `CONFIG_BT_NIMBLE_ROLE_PERIPHERAL` to zero, but commented it out (since I use `#if defined()`). What I saw was an even bigger code reduction on the peripheral when commenting `CONFIG_BT_NIMBLE_ROLE_CENTRAL` out, compared to the first commit. 

This PR should be checked by you! I did my best to sort out what class is for central, what for peripheral, and which classes are generic. The same is true for methods and class variables in NimBLEDevice, but I am not a 100% certain of making the right choices...!!

h2zero:
Thanks, yes I will be checking this soon, I have a lot of things going right now I was just curious if it had an effect on the NimBLE core code, never tried it 😄. Glad to hear we can use that to maximum benefit.

Jeroen88:
Yes it is a big memory saver if you need to implement just a central or a peripheral, really worth the effort!!

h2zero:
I plan on it, too many things going on at the moment, not enough time to really focus.

h2zero:
Good work on this, works pretty well. There's a few tweaks needed, if you want I can do them, otherwise I'll post comments for you.

Jeroen88:
> if you want I can do them, otherwise I'll post comments for you

I am glad to help, but I think it is more efficient if you do it :)

h2zero:
Sounds good, if you could check the box to “Allow edits from maintainers” that would be easiest.

Jeroen88:
That was already ticked :)

h2zero:
@Jeroen88 I just pushed some edits to this, could you test it on your setup? Took the opportunity to cleanup the code a bit since almost all the files were open, sorry if that makes it hard to follow the changes.

Jeroen88:
I browsed the full diff, not a lot of real changes I saw. The cleanup though was necessary!

I will try to pull the changes in later

h2zero:
Not a lot of changes, just added the ability to have scanner only and advertiser only setup and made it compatible with IDF settings.

Jeroen88:
Very nice, thank you!

PR #42, opened 2020-05-12
EXPERIMENTAL: Making remote object class member functions 'const'

Jeroen88:
Making remote object functions const. With this PR pointers to intermediate objects in the chain to the (remote) characteristic or (remote) descriptor can be created and maintained to be used several times:

```
void demo() {
  BLEClient * pClient = BLEDevice::createClient();
  bool success = pClient->connect(BLEAddress("a4:c1:38:5d:ef:16"));
  if(!success) {
    BLEDevice::deleteClient(pClient);
    Serial.println("Failed to connect");
    return;
  }
  const BLERemoteService * pRemoteService = pClient->getService(BLEUUID(0xebe0ccb0, 0x7a0a, 0x4b0c, 0x8a1a6ff2997da3a6));
  if(pRemoteService == nullptr) {
    Serial.print("Failed to find our service UUID: ");
  } else {
    Serial.println(" - Found our service");
    // Obtain a reference to the characteristic in the service of the remote BLE server.
    const BLERemoteCharacteristic * pRemoteCharacteristic = pRemoteService->getCharacteristic(BLEUUID(0xebe0ccc1, 0x7a0a, 0x4b0c, 0x8a1a6ff2997da3a6));
    if(pRemoteCharacteristic == nullptr) {
      Serial.println("Failed to find our characteristic UUID");
    } else {
      Serial.println(" - Found our characteristic");
      if(pRemoteCharacteristic->canRead()) {
        Serial.println("Can read");
        const uint8_t * pData = (const uint8_t *)pRemoteCharacteristic->readValue().data();
        float temperature = (pData[0] | (pData[1] << 8)) * 0.01;
        uint8_t humidity = pData[2];
        Serial.printf("LYWSD03MMC temperature = %.2f : humidity = %u\n", temperature, humidity);
      }
      if(pRemoteCharacteristic->canNotify()) {
        Serial.println("Can notify");
      }
    }
  }
  pClient->disconnect();
  BLEDevice::deleteClient(pClient);
}
```

**Warning**:
- This PR isn't thoroughly tested
- some `this` pointers needed const_cast, mostly because they modify m_value. Since m_value's are always returned by value, I think this is acceptable

By-catch, my example program is ~800 program bytes smaller. Because more values are passed as a reference, the performance should also be better, but Idid not time it.

This PR is done for 'central' only, 'peripheral' needs to be done.

### DIFF
--- a/src/NimBLEAdvertisedDevice.cpp
+++ b/src/NimBLEAdvertisedDevice.cpp
@@ -54,7 +54,7 @@ NimBLEAdvertisedDevice::NimBLEAdvertisedDevice() {
  *
  * @return The address of the advertised device.
  */
-NimBLEAddress NimBLEAdvertisedDevice::getAddress() {
+const NimBLEAddress &NimBLEAdvertisedDevice::getAddress() const {
     return m_address;
 } // getAddress
 
@@ -67,7 +67,7 @@ NimBLEAddress NimBLEAdvertisedDevice::getAddress() {
  *
  * @return The appearance of the advertised device.
  */
-uint16_t NimBLEAdvertisedDevice::getAppearance() {
+uint16_t NimBLEAdvertisedDevice::getAppearance() const {
     return m_appearance;
 } // getAppearance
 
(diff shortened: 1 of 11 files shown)


h2zero:
I like it so far, it’s a good start to const correct the rest of the library which should be done. One thing I see is m_value and the semaphores are a great use case for `mutable` declaration.

h2zero:
Closing this for now as there has been many changes since, will keep this in mind in the future.

PR #44, opened 2020-05-15
Correct error in creation of NimBLEAddress from six bytes in a std::string

Jeroen88:
Add creation of a NimBLEAddress from a std::string of 6 exactly bytes (no leading or trailing zeroes allowed). Add creation of a NimBLEAddress from an empty string, also because this is used in error (without purpose I think) in the library in the form of an address with six zeroes `NimBLEAddress("\0\0\0\0\0\0")`, which in fact leads to a string of length zero and the log text `Invalid address ''`.

### DIFF
--- a/src/NimBLEAdvertisedDevice.h
+++ b/src/NimBLEAdvertisedDevice.h
@@ -95,7 +95,7 @@ private:
     bool m_haveTXPower;
 
 
-    NimBLEAddress  m_address = NimBLEAddress("\0\0\0\0\0\0");
+    NimBLEAddress  m_address = NimBLEAddress("\0\0\0\0\0\0", 6);
     uint8_t         m_advType;
     uint16_t        m_appearance;
     int             m_deviceType;
--- a/src/NimBLEAdvertisedDevice.h
+++ b/src/NimBLEAdvertisedDevice.h
@@ -95,7 +95,7 @@ private:
     bool m_haveTXPower;
 
 
-    NimBLEAddress  m_address = NimBLEAddress("\0\0\0\0\0\0", 6);
+    NimBLEAddress  m_address = NimBLEAddress("");
     uint8_t         m_advType;
     uint16_t        m_appearance;
     int             m_deviceType;
--- a/src/NimBLEClient.h
+++ b/src/NimBLEClient.h
@@ -73,7 +73,7 @@ private:
     bool                retrieveServices();  //Retrieve services from the server
 //    void                onHostReset();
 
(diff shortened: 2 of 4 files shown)


h2zero:
Thanks for this, those log messages were bothering me 😃

Jeroen88:
> Thanks for this

You're welcome!

Jeroen88:
Change NimBLEAddress("\0\0\0\0\0\0") into NimBLEAddress("\0\0\0\0\0\0", 6) to have the correct size of six (filled with zeroes). Without it the address has size() zero, see #48.

src/NimBLEAdvertisedDevice.h
```diff
@@ -95,7 +95,7 @@ class NimBLEAdvertisedDevice {
     bool m_haveTXPower;
 
 
-    NimBLEAddress  m_address = NimBLEAddress("\0\0\0\0\0\0");
+    NimBLEAddress  m_address = NimBLEAddress("\0\0\0\0\0\0", 6);
```
h2zero: This will not compile, but we can remove it and just send and empty string thanks to the other changes you made.

src/NimBLEClient.h
```diff
@@ -73,7 +73,7 @@ class NimBLEClient {
     bool                retrieveServices();  //Retrieve services from the server
 //    void                onHostReset();
 
-    NimBLEAddress    m_peerAddress = NimBLEAddress("\0\0\0\0\0\0");   // The BD address of the remote server.
+    NimBLEAddress    m_peerAddress = NimBLEAddress("\0\0\0\0\0\0", 6);   // The BD address of the remote server.
```
h2zero: Same here

h2zero:
@Jeroen88 Minor comments left, but looks good.

src/NimBLEAdvertisedDevice.h
Jeroen88: No that will not compile indeed. Too quick...! I wanted to use `NimBLEAddress(std::string("\0\0\0\0\0\0", 6))` but I changed it into `NimBLEAddress("")`

src/NimBLEClient.h
Jeroen88: Same solution

Jeroen88:
Trailing and leading zeroes _are_ allowed as long as the address is constructed from six byte `std::string`, so e.g. `NimBLEAddress(std::string("\0x00\0x01\0x02\0x02\0x01\0x00", 6))` is perfectly valid.

h2zero:
Yes that is valid, there just wasn't the `std::string` construction before 😄. Thanks for updating.

PR #45, opened 2020-05-15
NimBLEScan results are incomplete!

Jeroen88:
Add `NimBLEAdvertisedDevice` copy constructor and use it in `NimBLEScanResults::getDevice()`, because otherwise the `std::string` class member variables of the advertisedDevice is not returned correctly by `NimBLEScan::getResults()` and `NimBLEScan::start()`. Only the data from the `BLEAdvertisedDeviceCallbacks` (so on a per device basis) is correct. 
I _think_ (but I am not a 100% sure) the root cause is because `std::map<std::string, NimBLEAdvertisedDevice*> m_advertisedDevicesMap` in `NimBLEScanResults` is pointer based, and thus the `NimBLEAdvertisedDevice*` pointers are _copied_ and not _constructed_.

The `uint8_t *payload` pointer may still be dangling!

### DIFF
--- a/src/NimBLEAdvertisedDevice.cpp
+++ b/src/NimBLEAdvertisedDevice.cpp
@@ -50,6 +50,37 @@ NimBLEAdvertisedDevice::NimBLEAdvertisedDevice() {
 
 
 /**
+ * @brief Copy constructor
+ */
+NimBLEAdvertisedDevice::NimBLEAdvertisedDevice(const NimBLEAdvertisedDevice &rhs) {
+    m_address          = rhs.m_address;
+    m_advType          = rhs.m_advType;
+    m_appearance       = rhs.m_appearance;
+    m_deviceType       = rhs.m_deviceType;
+    m_manufacturerData = rhs.m_manufacturerData;
+    m_name             = rhs.m_name;
+    m_pScan            = rhs.m_pScan;
+    m_rssi             = rhs.m_rssi;
+    m_serviceUUIDs     = rhs.m_serviceUUIDs;
+    m_txPower          = rhs.m_txPower;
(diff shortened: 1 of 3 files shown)


h2zero:
Good catch, can't say I've used that method much. Any reason not to just return a pointer/reference there instead of copying?

Jeroen88:
Well I did not design the API! ;)

I added `NimBLEAdvertisedDevice *getDevice(const NimBLEAddress &address)` in PR #46, and this works correctly without the copy constructor. However, I think it is not correct to return non-const pointers internal structures that should not be modified by the program. But non-const pointers are returned everywhere in this library! If we make them const, as suggested in PR #42, the user program should update all pointer variables to be const also, so this is not without impact.

If we return a reference we should come up with a solution to a getDevice() out-of-bounds. Now a nullptr is returned, then some kind of empty device should be returned.

Speaking of returning references, I think it would be cool to return references everywhere instead of returning pointers. But then we should introduce also 'empty objects' for (intermediate) non-existent services or characteristics. Also some kind of 'check breadcrumb trail' should be introduced in each leaf of the `service.characteristic.descriptor` tree, for the user to check, if the retrieval of a value fails, until what level the path is valid. **This heavily changes the API!**

Jeroen88:
I did some additional tests and I think the copy constructor is not necessary. Closing for now.

h2zero:
Yes ideally we a const reference should be returned wherever internal structures are returned but that's a pretty big ask. I don't mind things as they are so much since it would be the API user that messes up the data if they change things, that;s on them 😃. The only other alternative is to return copies of the data, which I don't like, but is necessary sometimes.

PR #47, opened 2020-05-17
Do not take the semaphore on an infinite scan (with duration == 0)

Jeroen88:
Do not take the semaphore on an infinite scan (with duration == 0), because this will block the program. If you don't take it, also don't release it when m_duration == 0.

With this PR you can do:

```
#include <Arduino.h>
#include <NimBLEDevice.h>

void setup() {
  Serial.begin(115200);
  Serial.println("/n/nStarted");

  BLEDevice::init("ESP32");

  Serial.println("Start background scan forever...");

  static BLEScan *pBLEScan = BLEDevice::getScan(); //create new scan
  pBLEScan->setActiveScan(false);
  pBLEScan->setInterval(256);
  pBLEScan->setWindow(255);
  pBLEScan->start(0);
}


void loop() {  
  Serial.println("INSPECT scanned device...");

  BLEScan *pBLEScan = BLEDevice::getScan(); // get pointer to existing scan in endlessly progress
  
  size_t index;
  bool found = false;
  for(index = 0; index < pBLEScan->getResults().getCount(); index++) {
    if(pBLEScan->getResults().getDevice(index).getAddress() == NimBLEAddress("58:2d:34:39:22:58")) {
      found = true;
      break;
    }
  }
  if(!found) {
    delay(12000);
    return;
  }

  NimBLEAdvertisedDevice dev = pBLEScan->getResults().getDevice(index);
  if(dev.haveServiceData()) {
    const uint8_t * pData = (const uint8_t*)(dev.getServiceData().data());
    Serial.printf("==== Data for uuid %s\n", std::string(dev.getServiceDataUUID()).c_str());
    size_t length = dev.getServiceData().length();
    for(size_t i = 0; i < length; i++) Serial.printf("0x%02x, ", pData[i]);
    Serial.println();
    if(pData[11] == 0x0d) {
      uint16_t rawTemperature = (pData[15] << 8) + uint16_t(pData[14]);
      uint16_t rawHumidity = (pData[17] << 8) + pData[16];
      Serial.printf("MJ-HT-V1 temperature is %.01f, humidity %.01f\n", float(rawTemperature) / 10.0, float(rawHumidity) / 10.0);
    } else if(pData[11] == 0x0a) {
      Serial.printf("MJ-HT-V1 battery is %u\n", pData[14]);
    } else if(pData[11] == 0x04) {
      uint16_t rawTemperature = (pData[15] << 8) + uint16_t(pData[14]);          
      Serial.printf("MJ-HT-V1 temperature is %.01f\n", float(rawTemperature) / 10.0);
    } else if(pData[11] == 0x06) {
      uint16_t rawHumidity = (pData[15] << 8) + uint16_t(pData[14]);          
      Serial.printf("MJ-HT-V1 humidity is %.01f\n", float(rawHumidity) / 10.0); 
    } else {
      Serial.printf("MJ-HT-V1 unknown type %u\n", pData[11]);
    }
  }

  delay(12000);
}
```

So you can start an infinite scan and read your device's service data at any time you want. You'll get the latest service data without the need for any callback function (in another PR yet in development I'll add a timestamp to the moment the advertised device was scanned).

I have not tested yet if this will co-exist with _reading_ a device. Maybe an infinite scan should be paused while reading and restarted afterwards. @h2zero what do you think?

### DIFF
--- a/src/NimBLEScan.cpp
+++ b/src/NimBLEScan.cpp
@@ -69,6 +69,7 @@ NimBLEScan::NimBLEScan() {
     m_scan_params.filter_duplicates  = 1; // If set, the controller ignores all but the first advertisement from each device.
     m_pAdvertisedDeviceCallbacks     = nullptr;
     m_stopped                        = true;
+    m_duration = 0;
     m_wantDuplicates                 = false;
 }
 
@@ -164,6 +165,7 @@ NimBLEScan::NimBLEScan() {
             }
 
             pScan->m_stopped = true;
+            pScan->m_duration = 0;
             pScan->m_semaphoreScanEnd.give();
             return 0;
         }
@@ -246,7 +248,9 @@ bool NimBLEScan::start(uint32_t duration, void (*scanCompleteCB)(NimBLEScanResul
     }
 
     m_stopped = false;
-    m_semaphoreScanEnd.take("start");
+    if(duration != 0) {
+        m_semaphoreScanEnd.take("start");
+    }
 
     // Save the callback to be invoked when the scan completes.
     m_scanCompleteCB = scanCompleteCB;
@@ -280,7 +284,10 @@ bool NimBLEScan::start(uint32_t duration, void (*scanCompleteCB)(NimBLEScanResul
         NIMBLE_LOGE(LOG_TAG, "Error initiating GAP discovery procedure; rc=%d, %s",
                                         rc, NimBLEUtils::returnCodeToString(rc));
         m_stopped = true;
-        m_semaphoreScanEnd.give();
+        if(m_duration != 0) {
+            m_duration = 0;
+            m_semaphoreScanEnd.give();
+        }
         return false;
     }
 
@@ -296,7 +303,9 @@ bool NimBLEScan::start(uint32_t duration, void (*scanCompleteCB)(NimBLEScanResul
  */
 NimBLEScanResults NimBLEScan::start(uint32_t duration, bool is_continue) {
     if(start(duration, nullptr, is_continue)) {
-        m_semaphoreScanEnd.wait("start");   // Wait for the semaphore to release.
+        if(duration != 0) {
+            m_semaphoreScanEnd.wait("start");   // Wait for the semaphore to release.
+        }
     }
     return m_scanResults;
 } // start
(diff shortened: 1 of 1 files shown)


h2zero:
Would it not work the same if you just called `scan->start(0,nullptr,false)`?

Jeroen88:
> Would it not work the same if you just called scan->start(0,nullptr,false)?

No this doesn't solve the problem

h2zero:
That's strange, it should work since that's basically what this PR does if duration == 0. It should not block unless calling scan again after the first time.

Jeroen88:
If duration == 0 the semaphore is never released, because [BLE_GAP_EVENT_DISC_COMPLETE](https://github.com/h2zero/NimBLE-Arduino/blob/9c011aa37402b60d6d06acc9581fe2e10ac00550/src/NimBLEScan.cpp#L158) is never issued.

h2zero:
It doesn't need to be released though unless there is something else trying to take it somewhere that I'm missing? I mean `scan->start()` can take that semaphore and keep it forever if it want's unless something else needs to use a resource it's controlling, but nothing else I can think of needs to take that resource, it was only for using the blocking call.

Jeroen88:
You are right (of course), I used the [second](https://github.com/h2zero/NimBLE-Arduino/blob/ec5b37cb98e9ae983d481fe2940f8f31732fbc35/src/NimBLEScan.cpp#L297) `start()` function, that indeed waits for the semaphore. Using `start(0,nullptr,false)` gives the behavior I need!. Closing this PR...

Jeroen88:
@h2zero Can an infinite scan co-exist with a read() or a write()? In other words, if I start an infinite scan, can I still read and write remote characteristics? Will notifications still be seen?

h2zero:
@Jeroen88 Yes you can read/write do whatever you like while scanning, the stack will just stop the scan to send/receive data from the device then resume the scan.

h2zero:
For what it's worth, this PR was a good idea. If someone called the blocking `start()` method with 0 duration their app would be stuck, unless they had another task that would stop it later manually. At the very least there should be a warning or something.

Jeroen88:
> For what it's worth, this PR was a good idea. 

Haha, after all a good idea! Would you like me to open it again? Do you need some changes, or will you finish it yourself?

Jeroen88:
> Yes you can read/write do whatever you like while scanning, the stack will just stop the scan to send/receive data from the device then resume the scan

It works like a charm!

h2zero:
> Lol, after all a good idea! Would you like me to open it again? Do you need some changes, or will you finish it yourself?

I think I may rework this a bit while I'm messing with the semaphores.

Jeroen88:
> I think I may rework this a bit while I'm messing with the semaphores

Okay, I'll keep it closed for now!

Jeroen88:
@h2zero I am cleaning up stale branches. Do we need this one still or can it be deleted?

h2zero:
You can delete this. I will be reworking this soon.

PR #46, opened 2020-05-17
Map to vector v2 (this was PR #37)

Jeroen88:
Add `NimBLEAdvertisedDevice *getDevice(const NimBLEAddress &address)` to get a pointer to an advertised device by address.

This is useful for devices that send information in their service data. Now you can do:
```
NimBLEAdvertisedDevice *dev = pBLEScan->getResults().getDevice(NimBLEAddress("58:2d:34:39:22:58"));
if(dev != nullptr && dev->haveServiceData()) {
  const uint8_t * pData = (const uint8_t*)(dev->getServiceData().data());
  ... do something with the service data ...
}
```

This was PR #37, but since @h2zero merged master into this PR at May 15, 2020, I could not push my branch again. So I created a new branch by pulling from PR #37. So this branch **includes** the master merged by @h2zero. I am not aware if this leads to additional diffs from that master with PR #37 ...

PR #49, opened 2020-05-17
Update NimBLERemoteCharacteristic value from a notification (replaces PR #27)

Jeroen88:
This PR:
- Updates the `m_value` in the `NimBLERemoteCharacteristic` from the stack callback
- Adds `std::string NimBLERemoteCharacteristic::getValue(time_t *timestamp = nullptr)` to access the value that is present in the characteristic _without reading it again_ from the remote server!
- Adds a timestamp (epoch) to the `NimBLERemoteCharacteristic`. The timestamp is set after a `::readValue()` and after a notification.
- Adds a new callback `void notifyCallbackPlain(BLERemoteCharacteristic* pBLERemoteCharacteristic, bool isNotify)`. In this new callback the `uint8_t *` pointer to the data and the length parameters are omitted, because they can be read from the characteristic itself, by `pBLERemoteCharacteristic->getValue()`.
- Adds a timestamp (epoch) to `NimBLEadvertisedDevice`. The timestamp is set each time the device is encountered during a scan
- Adds `NimBLEadvertisedDevice::getTimestamp()` to access the last time the device was scanned

With this PR:
- It is possible to read the service data and / or the manufacturer data without setting up callbacks for it, and without waiting for a scan to complete. This is especially useful in an infinite scan (with `duration == 0`). The callback still keeps on having the same behavior.
- It is also possible to get the value from a notification, again without setting up callbacks for it. Still the callbacks work properly. Moreover: In the notify callback the value can (also) be accessed from the notification itself.
- The API stays backward compatible, `remoteCharacteristic::readValue()` (and the other read functions) keep their purpose and work like before. The new `::getValue()` returns the latest value set either by `::readValue()` or by notification

This PR replaces PR #27. As stated by @h2zero in that PR, the device for what I created that PR, the LYWSD03MMC temperature / moisture sensor, does not comply with the BLE standards: it starts sending notifications on service ebe0ccb0-7a0a-4b0c-8a1a-6ff2997da3a6 characteristic ebe0ccc1-7a0a-4b0c-8a1a-6ff2997da3a6 right after connecting with it, without asking for that notification. With this PR you can ask that device for a notification on the service and characteristic, and even read it, and register a callback for it. That will work conform BLE standard (although connecting to the device is enough to get the notifications). In the notification callback you can read the data from the `uint8_t *` pointer OR from the characteristic itself using `getValue()`. Or you can access the latest value at any time by getting a reference to the characteristic and issue a `getValue()`

### DIFF
--- a/src/NimBLERemoteCharacteristic.cpp
+++ b/src/NimBLERemoteCharacteristic.cpp
@@ -418,6 +418,7 @@ int NimBLERemoteCharacteristic::onReadCB(uint16_t conn_handle,
             NIMBLE_LOGD(LOG_TAG, "Got %d bytes", attr->om->om_len);
 
             characteristic->m_value += std::string((char*) attr->om->om_data, attr->om->om_len);
+            characteristic->m_timestamp = time(nullptr);
             return 0;
         }
     }
--- a/src/NimBLERemoteCharacteristic.cpp
+++ b/src/NimBLERemoteCharacteristic.cpp
@@ -453,7 +453,7 @@ std::string NimBLERemoteCharacteristic::getValue(time_t *timestamp) {
     m_semaphoreReadCharEvt.take("readValue");
     std::string value = m_value;
     *timestamp = m_timestamp;
-    m_semaphoreReadCharEvt.give(1);
+    m_semaphoreReadCharEvt.give();
     return value;
 }
 
--- a/src/NimBLERemoteCharacteristic.cpp
+++ b/src/NimBLERemoteCharacteristic.cpp
@@ -56,6 +56,7 @@ static const char* LOG_TAG = "NimBLERemoteCharacteristic";
     m_charProp       = chr->properties;
     m_pRemoteService = pRemoteService;
     m_notifyCallback = nullptr;
+    m_notifyCallbackPlain = nullptr;
     m_rawData        = nullptr;
     m_dataLen        = 0;
     m_timestamp      = 0;
--- a/src/NimBLERemoteCharacteristic.cpp
+++ b/src/NimBLERemoteCharacteristic.cpp
@@ -390,8 +390,6 @@ std::string NimBLERemoteCharacteristic::readValue() {
 
     int rc = 0;
     int retryCount = 1;
-    // Clear the value before reading.
-    m_value = "";
 
     NimBLEClient* pClient = getRemoteService()->getClient();
 
@@ -403,6 +401,8 @@ std::string NimBLERemoteCharacteristic::readValue() {
 
     do {
         m_semaphoreReadCharEvt.take("readValue");
+        // Clear the value before reading.
+        m_value = "";
 
         rc = ble_gattc_read_long(pClient->getConnId(), m_handle, 0,
                                  NimBLERemoteCharacteristic::onReadCB,
@@ -446,13 +446,11 @@ std::string NimBLERemoteCharacteristic::readValue() {
  * @return The value of the remote characteristic.
  */
 std::string NimBLERemoteCharacteristic::getValue(time_t *timestamp) {
-    if(timestamp == nullptr) {
-        return m_value;
-    }
-
-    m_semaphoreReadCharEvt.take("readValue");
+    m_semaphoreReadCharEvt.take("getValue");
     std::string value = m_value;
-    *timestamp = m_timestamp;
+    if(timestamp != nullptr) {
+        *timestamp = m_timestamp;
+    }
     m_semaphoreReadCharEvt.give();
     return value;
 }
--- a/src/NimBLERemoteCharacteristic.cpp
+++ b/src/NimBLERemoteCharacteristic.cpp
@@ -480,10 +480,10 @@ bool NimBLERemoteCharacteristic::registerForNotify(notify_callback notifyCallbac
 
 
 /**
- * @brief Delete the descriptors in the descriptor map.
- * We maintain a map called m_descriptorMap that contains pointers to BLERemoteDescriptors
- * object references.  Since we allocated these in this class, we are also responsible for deleteing
- * them.  This method does just that.
+ * @brief Delete the descriptors in the descriptor vector.
+ * We maintain a map called m_descriptorVector that contains pointers to BLERemoteDescriptors
+ * object references. Since we allocated these in this class, we are also responsible for deleting
+ * them. This method does just that.
  * @return N/A.
  */
 void NimBLERemoteCharacteristic::removeDescriptors() {
--- a/src/NimBLERemoteCharacteristic.cpp
+++ b/src/NimBLERemoteCharacteristic.cpp
@@ -481,7 +481,7 @@ bool NimBLERemoteCharacteristic::registerForNotify(notify_callback notifyCallbac
 
 /**
  * @brief Delete the descriptors in the descriptor vector.
- * We maintain a map called m_descriptorVector that contains pointers to BLERemoteDescriptors
+ * We maintain a vector called m_descriptorVector that contains pointers to BLERemoteDescriptors
  * object references. Since we allocated these in this class, we are also responsible for deleting
  * them. This method does just that.
  * @return N/A.
--- a/src/NimBLERemoteCharacteristic.cpp
+++ b/src/NimBLERemoteCharacteristic.cpp
@@ -58,6 +58,7 @@ static const char* LOG_TAG = "NimBLERemoteCharacteristic";
     m_notifyCallback = nullptr;
     m_rawData        = nullptr;
     m_dataLen        = 0;
+    m_timestamp      = 0;
  } // NimBLERemoteCharacteristic
 
 
@@ -378,6 +379,23 @@ std::string NimBLERemoteCharacteristic::readValue() {
 
 
 /**
+ * @brief Get the value of the remote characteristic.
+ * @return The value of the remote characteristic.
+ */
+std::string NimBLERemoteCharacteristic::getValue(time_t *timestamp) {
+    if(timestamp == nullptr) {
+        return m_value;
+    }
+
+    m_semaphoreReadCharEvt.take("readValue");
+    std::string value = m_value;
+    *timestamp = m_timestamp;
+    m_semaphoreReadCharEvt.give(1);
+    return value;
+}
+
+
+/**
  * @brief Callback for characteristic read operation.
  * @return 0 or error code.
  */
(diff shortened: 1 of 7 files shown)


h2zero:
Good work here, might be an issue with the semaphore due to upcoming changes with those but we can worry about that after. I wish I had the time you do to work on this lol, I have too many other things going on lately, only get to work on this in small increments. Looking forward to testing later.

Jeroen88:
> Good work here

Thank you!

> I wish I had the time you do to work on this

It would be nice if you find some time to merge #37 / #46, to change the maps into vectors. I can merge these small changes on my local master for the time being, but the other keep me from further improvements.

h2zero:
I'll try to merge #37 tonight, could you cleanup the commented out code?

Jeroen88:
I can't push #37 anymore, because you merged master. I can cleanup #46 instead. But... we live in a different timezone! So next opportunity will be tomorrow to clean things up...

Jeroen88:
@h2zero I updated this PR with the lastest master

h2zero:
Perfect, thanks!

h2zero:
@Jeroen88 Is the new callback necessary here? I mean other than saving a couple bytes on the stack it doesn't seem to provide any functionality that the original doesn't have.

Jeroen88:
No not necessary because it does not add new functionality. I introduced it to maybe replace the current callback in the future

h2zero:
What I think should be done here in this case I’d comment out the new callback related code. Then we can discuss if anything related to the current callback should be changed. The rest looks great but that part should probably have it’s own PR.

Jeroen88:
> What I think should be done here in this case I’d comment out the new callback related code. 

Removed it completely, because:

> Then we can discuss if anything related to the current callback should be changed.

Well the change is very easy. Because with this PR the value is updated in the remote characteristic itself, also the pointer to the data and the length of the data could be taken from the characteristic itself, that is passed by the callback. This is just a minor improvement, I agree, so I leave it up to you @h2zero if you want to keep it this way or change it.

src/NimBLERemoteCharacteristic.cpp
```diff
@@ -463,7 +482,7 @@ bool NimBLERemoteCharacteristic::registerForNotify(notify_callback notifyCallbac
 /**
  * @brief Delete the descriptors in the descriptor vector.
  * We maintain a vector called m_descriptorVector that contains pointers to BLERemoteDescriptors
- * object references. Since we allocated these in this class, we are also responsible for deleteing
+ * object references. Since we allocated these in this class, we are also responsible for deleting
```
h2zero, line 529: You're good at finding these, I'm surprised you haven't seen the really bad one in client yet 😄.

h2zero:
> Well the change is very easy. Because with this PR the value is updated in the remote characteristic itself, also the pointer to the data and the length of the data could be taken from the characteristic itself, that is passed by the callback. This is just a minor improvement, I agree, so I leave it up to you @h2zero if you want to keep it this way or change it.

Yes there is nothing wrong with callback itself, it's just extra code to deal with is all. Would be fine if we were to replace the existing callback, but I'd rather not break the API unless it was a significant improvement.

src/NimBLERemoteCharacteristic.cpp
Jeroen88: Lol just by coincidence, and English isn't even my native language :smile:

Jeroen88:
@h2zero Updating to latest master

src/NimBLERemoteCharacteristic.cpp
```diff
@@ -440,6 +441,23 @@ std::string NimBLERemoteCharacteristic::readValue() {
...
+ */
+std::string NimBLERemoteCharacteristic::getValue(time_t *timestamp) {
+    if(timestamp == nullptr) {
+        return m_value;
+    }
+
+    m_semaphoreReadCharEvt.take("readValue");
+    std::string value = m_value;
+    *timestamp = m_timestamp;
+    m_semaphoreReadCharEvt.give(1);
```
h2zero: I know it seems strange but 0 == success in BLE (also 14 in NimBLE) anything else is an error value. So this should be give(0) or just give().

h2zero: Actually this method should probably be refactored a bit since we always return a copy of `m_value` .
Something like:
```
{
    m_semaphoreReadCharEvt.take("readValue");
    std::string value = m_value;
    m_semaphoreReadCharEvt.give();

    if(timestamp != nullptr) {
        *timestamp = m_timestamp;
    }

    return value;
}
```

Jeroen88: I wrote it like that on purpose. If we change it to your suggestion there is a little chance that the return value and the timestamp do not belong together, if a notification is in the middle of an update.

I do not like the extra copy of the value but I could not come up with another way. Do you?

h2zero: Making a copy is necessary but the compiler should optimize it to copy directly into the callers container. 

Also now that you mention the timing issue, we have no protection in the notification handler against the value being updated as `getValue()` is copying. The semaphore would only block if `readValue()` was called at the same time, but that would be odd.

Of course we cannot take a semaphore in the notification handler unless it was used with a 0 timeout then tested for success, but then we wouldn't be able to update the value if it failed.

Jeroen88: I am a semaphore zero  and you a h(2z)ero+e, what would you suggest?

h2zero: Considering the short term nature I think we could block in the notify event for a short time but it would require the use of it's own (new) semaphore to ensure that it would only block for the `getValue()` copy operation. 

So we could create `m_semaphoreNotifyEvt` then put `m_semaphoreNotifyEvt.take()` and `m_semaphoreNotify.give()` where it currently is in `getValue()` and then do the same [here](https://github.com/h2zero/NimBLE-Arduino/pull/49/files#diff-8f692c893f6a420d294d6426dcac9a9cR722). 

Would require stress testing but I don't think there will be any issues. The downside to this is an extra 200 bytes of memory per characteristic.

Jeroen88: Could you do it? I am really no good in semaphores...

Jeroen88: The 200 bytes is quite a lot, but I think it is worth it to have an characteristic value that is updated by notification. Will your dynamic semaphore trick make it any  better?

h2zero: I'll try it locally, if it works well I'll push the changes to this. 

Dynamic semaphore creation is a possibility here at the cost of speed for setup/alloc/dealloc, same as the other ones, but I think those can be replaced with task notifications.

Jeroen88: Thnx!

h2zero: I put this through a lot of stress testing for many hours. I used a loop calling `getValue()` every random 8-12us and a peripheral sending a notification every 50ms.

*  With no semaphore use at all there was no segfaults (wouldn't want to trust that, very odd).

* With static semaphore, no problem and appears safe.

* With dynamic semaphore, massive memory leaks due slow alloc/dealloc and not enough time for it to release the memory. Also a fault in the `take()` method as it took the semaphore while it was being deleted (at least this exposed a flaw in that idea and needs new thinking). 

Now for real world use I would consider the static semaphore route for the time being to stay on the safe side, if you agree I will push that addition to this PR. I hope for dynamic or another alternative to work in the future as the semaphores are expensive.

Jeroen88: You've put a lot of work in this, thank you so much for that!

> calling getValue() every random 8-12us and a peripheral sending a notification every 50ms

You are really pushing it to the limits!!

> With no semaphore use at all there was no segfaults (wouldn't want to trust that, very odd)

Can we come up with a reason for this? My conclusion was equal, although I did not call `getValue()` that frequently. Again, I am no semepahore expert, nor have I any knowledge of the ESP32 OS, but: could it be that something in FreeRTOS and/or the way that the running of programs and core functionality is organized on one or two cores prevents the occurrence of a segfault? Could it be that the semaphores are only necessary if the main programmer spawns more than one thread running BLE functions? 

> I would consider the static semaphore route for the time being to stay on the safe side

I agree with you, if we can't explain why segfaults aren't there, it is best to stay at the save side.

>  if you agree I will push that addition to this PR

Of course! I'd be very happy with that!

h2zero: >Can we come up with a reason for this?

My first thought is when we change the value in notify it creates a new string object and just changes the location pointer of m_value. If the copy constructor was called in `getValue()` at the same time it would still get a valid reference, at least until the original m_value destructor is called, which could be long enough to finish the copy. I could be completely wrong, worth testing though. 

The other possibility is flawed testing 😄, I just realized the peripheral did not vary the data length so my test never went out of range.

Jeroen88: > If the copy constructor was called in getValue() at the same time it would still get a valid reference, at least until the original m_value destructor is called, which could be long enough to finish the copy

I do not know how FreeRTOS switches tasks. We're not interrupt based, interrupts may occur everywhere. But maybe the OS has some mechanism to protect for the segfaults? idk...

h2zero: As far as I know it does not have any such mechanism. It simply switches between tasks on a priority basis, when one task is done or waiting for an event it switches to the next highest priority task. Interrupts are handled as their own, highest priority, task that stops all others I believe, but it's been a while since I got into that.

h2zero: I have come up with a compromise for this, I'll push it to this PR right away, stress testing for a hour now without a problem. It uses the same semaphore with a 0 timeout in the notify callback, that does not update the value if it fails to take it, but that should only happen when reading directly at the same time, so it should not be an issue 99.99% of the time. 

If you could test it later on your setup and let me know I'd really appreciate it.

Jeroen88: > If you could test it later on your setup and let me know I'd really appreciate it.

Sure! But most likely not today because I am busy at the office. Tomorrow for sure!

src/NimBLERemoteCharacteristic.cpp
```diff
@@ -402,6 +401,8 @@ std::string NimBLERemoteCharacteristic::readValue() {
 
     do {
         m_semaphoreReadCharEvt.take("readValue");
+        // Clear the value before reading.
+        m_value = "";
```
Jeroen88, line 405: Why do you clear m_value inside of the loop? Once should be enough. If you want it within the semaphore protection, you could also clear it inside the `if (rc != 0)`-block.

h2zero: I did that to protect it with the semaphore. The loop will only execute twice at most so it’s a minor issue.

If we put it inside the `if (rc != 0)` block it would only execute if there was an error in the call to `ble_gattc_read_long()`

Jeroen88: Looking at the [onReadCB](https://github.com/h2zero/NimBLE-Arduino/blob/92c0ca4630c519b1a4f010f9ef3971068aec1dc0/src/NimBLERemoteCharacteristic.cpp#L463) function now, that will always return zero. So in my suggestion it will never bet set to "".
But also looking at the CB, I think the value should NOT be cleared if, e.g. the CB is not meant for this connection (the first `if` in the CB). Although I am not sure, either the value should be kept to the existing value, or it should be set with a new, read, value, in all cases. But never cleared.

Jeroen88:
Not really a stress test, but getting the value updated from the notification every 12s for half an hour now without any problems!

src/NimBLERemoteCharacteristic.cpp
h2zero: It needs to be cleared before the call to `ble_gattc_read_long()`, when that is called it transmits the read request to the peripheral and when it gets a response the `onReadCB()` is triggered. In that callback we need to check if the connection id matches so we don't update the wrong characteristic, the `return 0` there just tells NimBLE to continue with the operation. After that we need to `m_value +=` to handle long reads that call the callback multiple times so it needs an empty container on the first call. Also if there was an error reading the value then it makes sense for it to be NULL since no data was read.

Jeroen88: Thnx for the explanation, as I understand it now `ble_gattc_read_long()` will generate one or _more_ CBs, and in each CB the `m_value` is updated by extension (`+=`). Still I think that it is impossible that `m_value` has been changed if the loop is entered a second time, but this is too minor to discuss about!

Jeroen88:
Is any action from my side needed or will you push this PR?

h2zero:
Nothing you need to do unless you wanted to change/add something else. Otherwise I will merge shortly.

Jeroen88:
No further plans!

PR #53, opened 2020-05-20
Create semaphores dynamically

h2zero:
This introduces dynamic creation of the semaphores used internally by the library, most noticeably in the client code (server/scan to be done). The reason for this change is each semaphore used in each attribute that uses them consumes ~180 bytes of memory. 

On the client especially, this effect is dramatic when considering there are 2 semaphores per remote service, 3 per remote characteristic and 2 per remote descriptor. If the remote device has 1 service with 1 characteristic and 1 descriptor, when we connect and obtain the database it will use ~1260 bytes of heap memory before beginning any other operations. 

This PR will remove that and create the semaphores on an as required basis using new macros to help implement this procedure. Using this method has reduced heap use in my testing by ~700% per connection, depending on the number of attributes on the remote device.

This is not without a cost however, it requires careful and strict adherence to properly creating and making sure the semaphores are deleted when no longer needed. In other words increased maintenance and increased risk of memory leaks and bugs. 

So far I have been able to run this iteration for days with 3 threads accessing the semaphore protected resources with potentially conflicting timings in both debug and release configuration without issue.

I believe this to be overwhelmingly worth the cost, @Jeroen88 has seen a ~27k heap use reduction connecting to a peripheral with 72 attributes, that is incredible and worth the effort IMHO.

If you can find a way to break this please comment, so far so good for me!

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,29 @@ A fork of the NimBLE stack restructured for compilation in the Ardruino IDE with
 
 Why? Because the Bluedroid library is too bulky. 
 
-Initial client code testing has resulted in code size reduction of ~115k and reduced ram consumption of ~37k.
+Initial testing has resulted in code size reduction of ~50% and reduced ram consumption of ~100k.
 
-Server code testing results from @beegee-toyo [from the project here](https://github.com/beegee-tokyo/ESP32WiFiBLE-NimBLE):
+## BLE_client example comparison (Debug):
+#### Arduino BLE Library   
+Sketch uses 1214909 bytes (57%) of program storage space.   
+Memory after connection: Free Heap: 171308   
 
+#### NimBLE-Arduino library
+Sketch uses 622076 bytes (29%) of program storage space.   
+Memory after connection: Free Heap: 270408   
+  
+## BLE_notify (server) example comparison (Debug):   
+#### Arduino BLE Library
+Sketch uses 1208409 bytes (57%) of program storage space.   
+Memory after connection: Free Heap: 173300   
 
+#### NimBLE-Arduino library   
+Sketch uses 609044 bytes (29%) of program storage space.   
+Memory after connection: Free Heap: 269448  
+   
+   
+## Server code testing results from @beegee-toyo [from the project here](https://github.com/beegee-tokyo/ESP32WiFiBLE-NimBLE):
+   
 ### Memory usage (compilation output)
 #### Arduino BLE library
 ```log
@@ -35,9 +53,8 @@ Flash: [=======   ]  69.5% (used 911378 bytes from 1310720 bytes)
 #### Arduino BLE library
 **`Internal Total heap 259104, internal Free Heap 91660`**    
 #### NimBLE-Arduino library
-**`Internal Total heap 290288, internal Free Heap 182344`** 
-  
-  
+**`Internal Total heap 290288, internal Free Heap 182344`**   
+
 # Installation:
 
 Download as .zip and extract to Arduino/libraries folder, or in Arduino IDE from Sketch menu -> Include library -> Add .Zip library.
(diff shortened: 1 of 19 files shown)


Jeroen88:
@h2zero as I said before, I am not very experienced with semaphores. Currently my own test case is using PR #49 but PR #40 not yet, 3 peripherals, endlessly scanned, connect (once) to one and maintain that connection. Next, every 12s: reading service data from the first, reading manufacturer data from the second, reading one characteristic from the third (connected) device. Will it give you insights if I start this scenario as a test for a longer period of time, to see if it is stable?

Jeroen88:
@h2zero I think I have to compliment you on this achievement, but frankly I have got no idea of what's going on, lol!

Jeroen88:
@h2zero I am thinking a bit about this PR. I do not know for sure, but if we merge PR #40, then most of the class instances for the characteristics and most of the descriptors will never be created. In this scenario the savings of this PR are far less than you report now. If we find a solution for creating services as well dynamically then also the service class instances are sparsely created.

IMHO dynamic population of the class instances is very important!

If we go on with the dynamic object creation, is the added complexity (and the, I think, non-standard use of semaphores which creates an extra threshold for other programmers that want to add to this library) of the dynamic semaphores worth the effort?
I am aware of the fact that in the dynamic objects scenario combined with the dynamic semaphores the memory savings increase on the number of objects requested by the programmer. Still I am in doubt of this solution....

h2zero:
Ideally, we could achieve both. Creating semaphores dynamically is quite common actually, they just sit there taking up resources most of the time, especially in this library. The maintenance isn't that bad, just like any other resource you allocate, you need to release it properly, the macros help with this. 

The complexity here comes from the use of a single static pointer to the allocated semaphore, which is a binary type wrapped in a class to make it somewhat custom and keeps track of the number of tasks trying to take it. This has been hidden in the class/macros and could probably be a bit simpler but I can see it being confusing for those who are unfamiliar with the subject. Perhaps I should explain the operation of them somewhere. 

#40 Is the next step, it's also a bit of a complex problem due to how BLE spec uses "handles" and how NimBLE accesses the peripheral database. I need to do some experiments with it and see what's reasonably possible. The question I need to answer there first is "does NimBLE request the whole database when calling get_service/characteristic_uuid?". If that is what it does we will have to implement some other way to get "only" the attributes we want.

Jeroen88:
> The question I need to answer there first is "does NimBLE request the whole database when calling get_service/characteristic_uuid?"

Do you mean the stack or the library? For the stack I don't know, but I can't imagine it retrieving more than needed. For the library I am sure: all services are retrieved, and next every characteristic for every service and every descriptor for each characteristic. Actually it is done with the function call [here](https://github.com/h2zero/NimBLE-Arduino/blob/a8f6e244e13559df33df6ae44c637ca43b231e51/src/NimBLEClient.cpp#L187) defined [here](https://github.com/h2zero/NimBLE-Arduino/blob/a8f6e244e13559df33df6ae44c637ca43b231e51/src/NimBLEClient.cpp#L400).

PR #40 retrieves all services on connection, but the characteristics dynamically only on first request. It does not retrieve the descriptors (yet).

h2zero:
I was referring to the stack, I'm not sure how it is done there. BLE was meant that you would connect and bond with the device so you would only ever get the database once and store a copy for future reconnection. So that is (somewhat) what the library does. 

Because of the way handles work you need to ask the device for the whole thing to find what you're looking for, it could be stopped once you found it but if the attribute you want is at the end of the database then you've just discovered the whole thing anyway. Also if you didn't store all those results and wanted a second attribute handle you would have to start the whole process over, wasting time and battery.

Jeroen88:
Moving the discussion to #40 ... see there

h2zero:
Thinking about this PR, even though it works well to my own surprise I believe there is a better option that can eliminate the semaphores completely. It’ll require a bit of work and testing but the results could be worth it.

h2zero:
Closing this as there is a better solution almost complete.

PR #54, opened 2020-05-20
Add iterators

Jeroen88:
Add iterators for `NimBLEScan`: `NimBLEadvertisedDevice`, `NimBLEClient`: `NimBLERemoteService`, `NimBLERemoteService`: `NimBLERemoteCharacteristic` and `NimBLERemoteCharacteristic`: `NimBLERemoteDescriptor`

This is handy e.g. for showing every address of the advertised devices from a scan. To do so, first get a new scan and next:

```
for(auto pAdvertisedDevice: pBLEScan->getResults()) {
  Serial.printf("Address is %s\n", std::string(pAdvertisedDevice->getAddress()).c_str());
}
```
Of course any other property of the advertised device can be shown (or looked up, if that is your use case)

Also this is handy e.g. for showing every UUID in a peripheral. To do so, first connect to a peripheral and next:

```
for(auto pService: *pClient) {
  Serial.printf("Service UUID is %s\n", std::string(pService->getUUID()).c_str());
  for(auto pCharacteristic: *pService) {
    Serial.printf("Characteristic UUID is %s\n", std::string(pCharacteristic->getUUID()).c_str());
    for(auto pDescriptor: *pCharacteristic) {
      Serial.printf("Descriptor UUID is %s\n", std::string(pDescriptor->getUUID()).c_str());
    }
  }
}
```
Again of course any other property can be shown, or looked up.

### DIFF
--- a/src/NimBLEClient.cpp
+++ b/src/NimBLEClient.cpp
@@ -349,6 +349,24 @@ int NimBLEClient::getRssi() {
 
 
 /**
+ * @brief Get iterator to the beginning of the vector of remote service pointers.
+ * @return An iterator to the beginning of the vector of remote service pointers.
+ */
+std::vector<NimBLERemoteService*>::iterator NimBLEClient::begin() {
+    return m_servicesVector.begin();
+}
+
+
+/**
+ * @brief Get iterator to the end of the vector of remote service pointers.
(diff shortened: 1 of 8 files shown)


h2zero:
Good QOL improvement for the application. Looks good to me.

PR #59, opened 2020-05-22
Slightly refactoring vector release functions and adding new ones by UUID

Jeroen88:
Make vector release functions public. Rename all functions to 'clear' for consistency. Add clear functions per UUID.

With this PR all services or just one service can be released from the service vector. The same is available for characteristics and descriptors. This comes in handy if e.g. you want to connect to a device, inspect some static value (e.g. from Generic Access, 0x1800), release that service to free the allocated memory for it, and go on with reading more volatile data from other services.

- The functions for releasing all objects where already in the library, but `private` and not consistently named. These are all `public` now and renamed to `clear()`. This can be done without breaking the API because they were private. 
- The functions for releasing one object by UUID are added and called `clear(const NimBLEUUID &uuid)`.
- In `NimBLEScan` the public functions `clearResults()` and `erase(const NimBLEAddress &)` are similar and already `public`. Because they are `public`, they are not renamed to 'clear' for not breaking te API.

The new functions are:
Release all services with all its characteristics and all its descriptors:
``
pClient->clear();
``

Release one service by UUID and all characteristics and descriptors associated with that service:
``
pClient->clear(NimBLEUUID(uint16_t(0x180a)));
``

Release all characteristics and all descriptors of one service:
``
pClient->getService(NimBLEUUID(uint16_t(0x180a)))->clear();
``

Release one characteristic by UUID and all descriptors associated with that characteristic:
``
pClient->getService(NimBLEUUID(uint16_t(0x180a)))->clear(NimBLEUUID(uint16_t(0x2a26)));
``

Release all descriptors of one characteristic:
``
pClient->getService(NimBLEUUID(0xebe0ccb0, 0x7a0a, 0x4b0c, 0x8a1a6ff2997da3a6))->getCharacteristic(NimBLEUUID(0xebe0ccc1, 0x7a0a, 0x4b0c, 0x8a1a6ff2997da3a6))->clear();
``

Release one descriptor by UUID 
``
pClient->getService(NimBLEUUID(0xebe0ccb0, 0x7a0a, 0x4b0c, 0x8a1a6ff2997da3a6))->getCharacteristic(NimBLEUUID(0xebe0ccc1, 0x7a0a, 0x4b0c, 0x8a1a6ff2997da3a6))->clear(NimBLEUUID(uint16_t(0x2901)));
``

The latter two functions are there for consistency and completeness, but I expect that they will be rarely used.


h2zero:
This is good, we should talk about some overloaded functions for this and for the get() operations. 

I’m thinking about a user supplied vector or array of uuids they want to keep or retrieve so if they wanted 2 out of 5 characteristics of a service for instance they could supply those in 1 call that would either retrieve from get() or save from clear().

Thoughts?

Jeroen88:
> a user supplied vector or array of uuids they want to keep or retrieve 

At first I was thinking in that direction too, but I let go of that thought. Thinking about real life users and real life applications, I think there are two or three broad use cases:
1. One doesn't know a device and wants to explore it: just retrieve everything that is there
2. One wants to read a few values and the UUID's in the path are know: just get it
3. Same as 2. but static data is read only once. Most of the time this will be from one service. Read that service and clear it afterwards

I think especially the `client::clear()` (for a service) fits use case 3. The others are a bit of an overkill I think but I made them for uniformity and completeness.

Speaking about these kind of functions, `getService()`, `getCharacteristic()` and `getDescriptor()` could all be renamed to just `get()`. What they will get is perfectly clear from the context. It saves the programmer a lot of typing, but breaks the original API.
Also, I think I will add `readValue<T>()` today, to be consistent with `getValue<T>()` and add the possibility to read other data types than `std::string`, `uint8_t`, `uint16_t` and `uint32_t`. Next I would make the `readUint8()`, `readUint16()` and `readUint32()` functions deprecated, but what would you do?

Jeroen88:
@h2zero Updating to latest master. I used the function name` clear()` everywhere, you used `remove...()` everwhere. In this commit all functions are named `clear()`, but e.g. `removeCharacteristic(UUID)` and `removeCharacteristics()`  etc. is more in line with `getCharacteristic(UUID)` and `getCharacteristics()`.

h2zero:
As much as I dislike typing all that too, I think the function names should stay for readability. For instance `characteristic->clear()` is ambiguous, the user may believe that it is clearing everything from that object such as values or other data which is not what is happening. Same goes for the proposed `get()` shortcuts, it's too ambiguous, rather the call should describe the intent of what they want to get. 

Believe me, I'd love to have my application be able to have one line of code: `getDataAndDoStuff()` but that wouldn't help the next person that has to work on it lol (I think I'll sneak that one into the library 😄). 

That said I like where this is going and the functionality should be there for the user if they want it.

>Also in your suggestion the programmer first has to add a number of specific UUIDs to a vector and then call clear() once. In the current scenario, the programmer calls clear() for each UUID. I think the latter is a little bit more intuitive.

Or we could make it easy and use a list initializer
```
void clearExcept(const std::list<std::string> & uuids) {
    for(auto chr : getCharacteristics()) {
        auto it = std::find(uuids.begin(), uuids.end(), std::sting(chr->getUUID()));
        if(it == uuids.end()) {
            delete chr;
            m_characteristicVector.erase(chr);
        }
}

pService->clearExcept({"1234","ABCD"});
```
Very quick and dirty, just spitballing, but something like that would be a nice addition. This way if there is a device with a service that has 10 characteristics and we only want, say the temperature and battery characteristic we could remove all but those in a single call rather than looping through manually deleting what we don't want, or deleting all and re-creating just those.

Jeroen88:
I renamed the functions back to `delete...()`, with ... equals Service(-s), Characteristic(-s) or Descriptor(-s), depending on what it deletes. I thought 'delete' was a better option, because it actually _deletes_ the object, not only _removes_ it from the vector.

> As much as I dislike typing all that too, I think the function names should stay for readability

Actually the real reason for this is I wrote 'ch@rtr!stic$' a million times since I work on this library, not once without a typo.

> `getDataAndDoStuff()`

I am really looking forward to that new functionality.

> Or we could make it easy and use a list initializer

That is indeed an elegant option. I am still not convinced however: If you want to delete e.g. more than one service, you first have to do something with that service, otherwise you wouldn't have to retrieve that service in the first place. Something like:

```
BLEUUID uuid = BLEUUID(first UUID);
BLEService *pService = pClient->getService(uuid);
... do something with the service
pClient->deleteService(uuid);
uuid = BLEUUID(next UUID);
pService = pClient->getService(uuid);
... do something with the service
pClient->deleteService(uuid);
```

I think the list initialized delete functions are a bit of an overkill. I think the delete functions as they are there now already will not be used very much. The list initialized delete functions even less.
I leave it up to you if you want to create them! :grin:. Take care though, your quick and dirty code is incorrect: you are looping through a vector while deleting elements of it!

h2zero:
> Take care though, your quick and dirty code is incorrect: you are looping through a vector while deleting elements of it!

Lol yeah, that was the “dirty” part.

But you’re right I don’t see a lot of use for this except when we have retrieved the whole database from the sever and want to trim it back to a few attributes.

PR #60, opened 2020-05-22
Remove automatic discovery in NimBLEClient::connect().

h2zero:
Instead of discovering the peripheral database on connection and consuming the associated resources this will give the user more control over the discovery operation.

* Adds `void NimBLEClient::discoverAtrributes()` for the user to discover all the peripheral attributes as a replacement for the former functionality.

* `getServices()`, `getCharacteristics()`, `getDescriptors()` now take an optional `bool` parameter (`default false`). If `true` it will clear the respective vector and retrieve all the respective attributes from the peripheral. If false it will retrieve the attributes only if the vector is empty, otherwise the vector is returned with the currently stored attributes.

* `getService(NimBLEUUID)`, `getCharacteristic(NimBLEUUID)`, `getDescriptor(NimBLEUUID)` will now  check the respective vectors for the attribute object and, if not found, will retrieve (only) the specified attribute from the peripheral.

User API functionality remains unchanged except for the added (optional) parameters stated above.

### DIFF
--- a/API_DIFFERENCES.md
+++ b/API_DIFFERENCES.md
@@ -122,6 +122,35 @@ Has been removed from the API as it is no longer maintained in the library.
 
 The last two above changes reduce the heap usage significantly with minimal application code adjustments.   
 
+**NEW** on May 23, 2020
+> ```
+> NimBLEClient::getServices(bool refresh = false)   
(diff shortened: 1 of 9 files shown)


Jeroen88:
It works like a charm :smiley: Thank you so much!!!

PR #64, opened 2020-05-27
Make variable static

Staars:
Otherwise this is visible to the outside and it has a dangerous short generic name. At least with PlatformIO this led to collision and a failed build.
The other driver did the same "bad" thing ;)


h2zero:
Thanks, I can certainly see this causing an issue.

PR #69, opened 2020-06-01
Add register for notify without callback

Jeroen88:
Since PR #49 it is possible to access the value set in a remote characteristic by a notification at any time _without using a callback function_ using `getValue()`. For this the possibility is added to register for a notification without a callback function, `registerForNotification(bool, bool)`.
Also in `registerForNotification()` an unneeded test is removed.

I choose for this option to not change the api. In the original design passing a nullptr to  `registerForNotification(bool, bool)` means unregister for notification. I think it would be better to change the order of the parameters and semanitics into `registerForNotification(bool notifications = true, bool response = true, notify_callback = nullptr)` with the semanitic of a nullptr being: _register_ for notification without a callback function instead of _unregister_ for notification, and add a `unregisterForNotification()` function. This will however change the API

### DIFF
--- a/src/NimBLERemoteCharacteristic.cpp
+++ b/src/NimBLERemoteCharacteristic.cpp
@@ -525,13 +525,13 @@ bool NimBLERemoteCharacteristic::setNotify(uint16_t val, bool response, notify_c
  * is performed for notifications.
  * @return true if successful.
  */
-bool NimBLERemoteCharacteristic::subscribeForNotify(bool notifications, bool response, notify_callback notifyCallback) {
+bool NimBLERemoteCharacteristic::subscribe(bool notifications, bool response, notify_callback notifyCallback) {
     if(notifications) {
         return setNotify(0x01, response, notifyCallback);
     } else {
         return setNotify(0x02, response, notifyCallback);
     }
-} // subscribeForNotify
+} // subscribe
 
 
 /**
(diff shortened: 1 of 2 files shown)


h2zero:
I really don't like how the register for notification setup currently works either, this is a good start to improving it, just needs the "unregister' part to go with it.

I think a new set of methods should be created, `subscribe(bool notifications = true, bool response = true, callback = nullptr)` and `unsubscribe()`  or something along those lines.

Jeroen88:
@h2zero I introduced `subscribeForNotify()` and `unsubscribeForNotify()` and a private helper function `setNotify()`. In `subscribeForNotify()` passing a NULL pointer as parameter now means: do subscribe for notifications (that can be read with `getValue()`) but don't call me back.

h2zero:
Looks good, I'd like to retain the original `registerForNotify()` intact and just add these on top of it. If you don't mind I can push the edits to this PR.

Jeroen88:
> If you don't mind I can push the edits to this PR

Please!

h2zero:
@Jeroen88 I pushed the changes, let me know what you think.

Jeroen88:
I have no possibilities to run the code until next Sunday, but reviewing it, it looks (very) good to me, including the nice use of the `subscribe()` and `unsubcribe()` in the deprecated function.

PR #70, opened 2020-06-03
Add uint64_t cast of an address

Jeroen88:
Now you can do

```
NimBLEAddress address("12:34:56:78:90:ab");
Serial.printf("Address is %12llx\n", uint64_t(address));

```
resulting in:

`Address is 1234567890ab`

### DIFF
--- a/src/NimBLEAddress.cpp
+++ b/src/NimBLEAddress.cpp
@@ -142,4 +142,10 @@ NimBLEAddress::operator std::string() const {
     return std::string(buffer);
 }
 
+NimBLEAddress::operator uint64_t() const {
+    uint64_t address = 0;
+    memcpy(&address, m_address, sizeof m_address);
+    return address;
+}
+
 #endif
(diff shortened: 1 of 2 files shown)


h2zero:
I'm not opposed to adding this but I'm curious as to why? I personally can't think of anything I would use it for, I'd like to see your use case.

Jeroen88:
>  personally can't think of anything I would use it for, I'd like to see your use case

Well I use it! 😄 

I read sensor values and send these to a cloud server. I want to uniquely identify the sensor and use the address for this. I'd rather send and store a uint64_t (8 bytes) than a MAC (18 bytes). Without a cast I would need to parse the MAC xx:xx:xx:xx:xx:xx. With cast this is easy.

h2zero:
That makes sense, you could also just use `NimBLEAddress::getNative()` and send/store those 6 bytes. I'll merge this soon, was just curious about it's use, thanks!

Jeroen88:
> I'll merge this soon

Thank you!

PR #72, opened 2020-06-07
Refactor server code 

h2zero:
This changes the server database to use vectors instead of maps for reduced resource use and better symmetry with the client code.

Implements critical sections on characteristic/descriptor value data to prevent concurrent access.

Fixes an issue when advertising without any services that would not advertise device name/txpower.


Jeroen88:
Nice! 😄

PR #73, opened 2020-06-11
Replace semaphores with task notifications

h2zero:
This completely removes all semaphores and uses task notifications and critical sections to replace their functionality.

The effect is a significant reduction in memory use (approx. 2kB/connection to devices with 6 attributes).  This also reduces flash size by a few kB and improves performance of the operations changed by nearly 50%.

 Also changed in this: 
- `NimBLEClient::connect()` will now stop scanning automatically when called.
-  minor bugfixes/cleanup


Jeroen88:
Very nice!

I am not experienced in tasks, so just curious: do task notifications have the same functionality as semaphores? No drawbacks? Will they work if the library was used in multitasking?

One question: any reason not to bring `getValue()` in line with `remoteDescriptor::getValue()` using  `template<typename T>`? Or has a descriptor always have the same type? If so, why not add e.g. `uint16_t getValue()` instead of `std::string getValue()`?

h2zero:
>I am not experienced in tasks, so just curious: do task notifications have the same functionality as semaphores? No drawbacks? Will they work if the library was used in multitasking?

They are similar but actually quite different. Semaphores will block any other task trying to take the semaphore, task notifications will not, they simply block when `take` (actually xTaskNotifyWait internally) is called and only block that task. No drawback if used correctly, as I believe I have done here, the stack calls should lock internally so concurrency risk. The semaphores were mainly used previously to block the tasks while waiting for the callback and to prevent concurrent data access, task notifications and critical sections proved to be a better choice for this purpose.

> One question: any reason not to bring getValue() in line with remoteDescriptor::getValue() using template<typename T>? Or has a descriptor always have the same type? If so, why not add e.g. uint16_t getValue() instead of std::string getValue()?

Simple answer; I forgot 😄. Descriptors can be used in the same way as characteristics if they are custom, they just don't have notification functionality. I'll add the templates later today, thanks for pointing that out.

Jeroen88:
As long as the deprecated functions are there, you could replace them with a template call, eg `readInt32()` can be as easy as `return readValue<int32_t>()`. Same is true for the other read functions.

Jeroen88:
NimBLERemoteService could use the template readValue() too.

I was questioning myself, why adding getValue() and a timestamp to NimBLERemoteDescriptor? For the NimBLERemoteCharacteristic it is introduced to be able to access a value set by a notification, and to know at what time it was set. But descriptors can only be read, so timekeeping is not necessary, and getValue() could be skipped, although keeping it offers a small benefit of not having to copy the data yourself if you need it several times.

h2zero:
> you could replace them with a template call

Yes, that should be done.

> why adding getValue() and a timestamp to NimBLERemoteDescriptor?

Only added for consistency with the same methods from NimBLERemoteCharacteristic. I contemplated removing the timestamp for descriptor and might still, but it doesn’t hurt anything to leave it.

h2zero:
> why adding getValue() and a timestamp to NimBLERemoteDescriptor?

>Only added for consistency with the same methods from NimBLERemoteCharacteristic. I contemplated removing the timestamp for descriptor and might still, but it doesn’t hurt anything to leave it.

Changed my mind and removed it all, no need to store the value of a descriptor, the application can do that if desired. Should be left for a separate commit if requested in the future anyway.

Jeroen88:
> no need to store the value of a descriptor, the application can do that if desired

Indeed! 😄

Issue #83, opened 2020-07-13
Possibility to identify the client accessing a characteristic


lknop:
I am in the process of migrating from bluedroid stack to NimBLE. However, there is one feature I have been unable to achieve in the old stack, and it seems that thanks to the feature parity, the same problem exists in NimBLE. 

Assuming BLEServer multiconnect, I need to determine the address of the device that is currently writing to the characteristic I exposed with BLEServer.

It seems that all that is needed is to pass the `conn_handle` parameter of `NimBLECharacteristic::handleGapEvent` in the `onWrite` callback. Is this correct or am I missing something?

I envision an approach analogous to `NimBLEServer` dual callback call:                  
```
server->m_pServerCallbacks->onConnect(server);
server->m_pServerCallbacks->onConnect(server, &desc); 
```

Since in the `onConnect` callback I can retrieve the `conn_handle` from `ble_gap_conn_desc` passed as `&desc` I could either keep the map of conn_handles and addresses, or use `ble_gap_conn_find` to find what the peer address for the given `conn_handle` is (assuming that the onWrite callback will receive the conn_handle as the second parameter).

Please be so kind to point any shortcomings of my reasoning, and if it should indeed work, let me know if I should create a PR for this.

h2zero:
Hi @lknop, yes this is certainly doable. 

It would be a simple matter to add the extra callback method and use `ble_gap_conn_find(conn_handle, &desc )` in the characteristic gap event handler, then calling the new method, just like it's done in the server gap event handler.

Adding this to the read event would be a good idea too. 

If you want to PR this, please feel free to do so. Otherwise I can add this when I get done with a few other things I'm working on. 

Thanks for the suggestion 😄 .

PR #85, opened 2020-07-14
Alternative characteristic read/write callbacks

lknop:
As mentioned in #83, I added extended read/write callbacks which pass the connection description information. Now I am able to retrieve the peer address at the write event.


h2zero:
Looks good to me! Thanks.

PR #91, opened 2020-07-20
Server: Add onSubscribe() callback for characteristics.

h2zero:
Adds a new method to NimBLECharacteristicCallbacks that gets called when a client changes it's subscription status.


h2zero:
As an addition to this I’m considering removing the descriptor callback for 2902 descriptors. 

Any feedback on do so would be appreciated. 

@mitchjs I know you make use of the 2902 callback, is this PR a suitable replacement for your purposes?

mitchjs:
@h2zero , I dont use of any 2902 descriptor callbacks... also not sure why one would need to.
i just use the standard characteristic callbacks... onread/onwrite/onnotify

h2zero:
@mitchjs sorry I thought I remembered you using the 2902 callbacks to detect when a client subscribed to your characteristic. Please disregard if I’m mistaken. Thanks for replying.

h2zero:
It seems this ha opened up a bigger hole. After removing the callbacks for the 0x2902 I came to realize the whole class is useless at that point... bigger changes to this PR incoming.

elguiri:
@h2zero I'll test later on, thanks. As an aside, there is still a typo in the PR title (subscibe not subscribe).

h2zero:
Lol I do that one often, thanks!

elguiri:
@h2zero I've had chance to test the code to date on this PR.

I was manually creating 2902 descriptors in my code eg. `characteristic->createDescriptor("2902", properties);`. This compiled fine, but then "crashed" my code with an assertion that "0x2902 descriptors cannot be manually created" as expected. I removed my manual creation (I was only doing that so I could use the onSubscribe callback that has now been moved to the characteristic).

The removal of the ability to manually create the 2902 descriptor will need to be documented accordingly, as it's a breaking change (and one that might not be obvious from a successful compilation).

I'm very happy with the onSubscribe on the characteristic, thanks very much for thinking about my underlying requirement!

h2zero:
Thanks for testing! I don’t see any issue removing the 2902 class. Any functionality it had could be easily incorporated into the characteristic if anyone needs it.

The documentation has already been changed in this PR :)

Issue #92, opened 2020-07-20
More examples?


h2zero:
I’m thinking this library could use a couple more examples. Does anyone have suggestions on what would be informative? 

I have some of my own ideas but I’d like to hear from users as to what they would like to see.

Also if anyone would like to submit a PR for any examples (or anything else) please feel free to do so.

alvarowolfx:
I'm creating a demo on how to provision a device (ESP32) to Google IoT Core over BLE. Basically I'm starting a BLE Uart interface when the user clicks a button and them the device can accept things like Wifi creds, IoT Core settings and private key.

https://github.com/alvarowolfx/gcloud-iot-core-device-provisioning

I can try to submit an example with just the Wifi provisioning part, I think that would be really valuable for users of this library.

BTW, awesome work that you're doing, saw major memory improvements here and is working flawlessly.

h2zero:
@alvarowolfx That's a nice project! 

> I can try to submit an example with just the Wifi provisioning part, I think that would be really valuable for users of this library.

That would be a great addition, if you have time to do it I would certainly include it.

txf-:
> I’m thinking this library could use a couple more examples. Does anyone have suggestions on what would be informative?

A bonding example would be useful. I'm currently having issues implementing automatic bonding on connection, such as what happens with the original ESP32 stack.

h2zero:
Hi @txf-, good suggestion. I am going to be improving that aspect of the library shortly to add more functionality and will create an example to show the use of it. 

For your purposes right now all you should need to do is call:
```
NimBLEDevice::setSecurityAuth(true, false, true);
```
The first parameter set to `true` enables bonding.

After that to replicate the original library functionality (for a server) you would need to call:
```
NimBLEDevice::startSecurity(uint16_t conn_id); 
```
From the `NimBLEServerCallbacks::onConnect(NimBLEServer* pServer, ble_gap_conn_desc* desc)` callback.
The `conn_id` is available from `desc-> conn_handle` 

For a client app you just need to call `NimBLEClient::secureConnection()` after connecting.

Sorry this is not described well, still a work in progress.

hannahvsawiuk:
Hi @h2zero, I am also trying to implement automatic bonding upon connection. The security settings were configured as you mentioned and custom server callbacks class was defined as such

```cpp
class MyServerCallbacks: public NimBLEServerCallbacks {
    void onConnect(NimBLEServer* pServer) {
[...]

PR #94, opened 2020-07-21
Add clearAll parameter to deinit(), Fixes #12.

### LINKED ISSUE #12: [Feature Request] Ability to clean up memory
In my situation I am only using ble to do setup, so after my device is set up I would like to be able to completely clean up the memory and move on. So it would be great if there was a method to clean up the singletons, or some way for me to delete.

I have found I can do most with client code, but the singleton objects are impossible to fix up without library changes.

Here is what I was thinking
```

void NimBLEDevice::cleanup() {
  if(NimBLEDevice::m_pServer == nullptr) {
      delete NimBLEDevice::m_pServer;
      NimBLEDevice::m_pServer = nullptr;
   }
  if(NimBLEDevice::m_bleAdvertising== nullptr) {
      delete NimBLEDevice::m_bleAdvertising;
      NimBLEDevice::m_bleAdvertising= nullptr;
   }
  if(NimBLEDevice::m_pScan== nullptr) {
      delete NimBLEDevice::m_pScan;
      NimBLEDevice::m_pScan= nullptr;
   }
   for(auto it = m_cList.cbegin(); it != m_cList.cend(); ++it)
   {
      deleteClient(*it);
   }
   // Whatever other clean up needs to be done.
  
}
```

h2zero:
Adds the ability to clear all resources consumed during BLE operation when deinitializing and shutting down BLE.
Useful when BLE is used intermittently or in a task that initializes, performs operations then deinitializes.

By setting the clearAll parameter to true all created BLE objects will be deleted, freeing the memory for other tasks.

Warning: This will invalidate any pointers that may be referencing the deleted objects.

### DIFF
--- a/src/NimBLEDevice.cpp
+++ b/src/NimBLEDevice.cpp
@@ -539,6 +539,8 @@ void NimBLEDevice::stopAdvertising() {
 
 /**
  * @brief Shutdown the NimBLE stack/controller.
+ * @param [in] clearAll If true, deletes all server/advertising/scan/client objects after deinitializing.
+ * @note If clearAll is true when called, any references to the created objects become invalid.
  */
 /* STATIC */ void NimBLEDevice::deinit(bool clearAll) {
     int ret = nimble_port_stop();
--- a/src/NimBLEDevice.cpp
+++ b/src/NimBLEDevice.cpp
@@ -540,7 +540,7 @@ void NimBLEDevice::stopAdvertising() {
 /**
  * @brief Shutdown the NimBLE stack/controller.
  */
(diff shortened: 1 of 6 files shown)


h2zero:
Sorry this took so long @schlaegerz. 

Server cleanup code had to be implemented first which is finally done.

alvarowolfx:
Awesome, I'm eager to try that out on the provisioning project that I mentioned on #92. I was having issues turning off the BLE service after things were done.

h2zero:
Yes I could see it being very useful for that. If you try it out let me know if there’s any issues.

Issue #98, opened 2020-07-25
[Feature proposal] Purpose of facilitating client reconnection process


wakwak-koba:
As I wrote the issue before, I mainly use BLEClient.
Issue at the time https://github.com/h2zero/NimBLE-Arduino/issues/7

In the process, it would be useful to have the following features.

## What I really want:

1. `uint8_t NimBLEClient::getPeerType()`
With `getPeerAddress`, I can get the most recently used BLE-Address(m_peerAddress), but can't get PeerType.
I want to implement m_peerType as well as m_peerAddress
```c++
uint8_t NimBLEClient::getPeerType() {
    return m_peerType;
} // getPeerType
```

2. Change the operation when the argument is omitted
    -`NimBLEClient::getServices(bool refresh = false)`
    -`NimBLERemoteService::getCharacteristics(bool refresh = false)`
    -`NimBLERemoteCharacteristic::getDescriptors(bool refresh = false)`
Although it is `(bool refresh = false)`, it is better to use `(bool refresh = true)` only for the first time.

```c++
std::vector<NimBLERemoteService*>* NimBLEClient::getServices() {
	return getServices(m_servicesVector.size() == 0);
} // getServices
```
```c++
std::vector<NimBLERemoteCharacteristic*>* NimBLERemoteService::getCharacteristics() {
    return getCharacteristics(m_characteristicVector.size() == 0);
} // getCharacteristics
```
```c++
std::vector<NimBLERemoteDescriptor*>* NimBLERemoteCharacteristic::getDescriptors() {
	return getDescriptors(m_descriptorVector.size() == 0);
} // getDescriptors
```

## What I want:

3. Specify the connection destination at the time of `NimBLEDevice::createClient()`
Since I can't connect during scan, I want to createClient in advance and connect after scan.
```c++
static NimBLEClient*    createClient(NimBLEAdvertisedDevice* device = nullptr);
```
```c++
/* STATIC */ NimBLEClient* NimBLEDevice::createClient(NimBLEAdvertisedDevice* device) {
    if(m_cList.size() >= NIMBLE_MAX_CONNECTIONS) {
        NIMBLE_LOGW("Number of clients exceeds Max connections. Max=(%d)",
                                            NIMBLE_MAX_CONNECTIONS);
    	return nullptr;
    }
	
    NimBLEClient* pClient;
    if(device == nullptr)
        pClient = new NimBLEClient();
    else
        pClient = new NimBLEClient(device);
    
[...]

PR #101, opened 2020-07-27
Add 'NimBLEClient::getPeerAddrType()'

wakwak-koba:
It is a case proposed in the following issue.
https://github.com/h2zero/NimBLE-Arduino/issues/98

### DIFF
--- a/src/NimBLEClient.cpp
+++ b/src/NimBLEClient.cpp
@@ -168,7 +168,7 @@ bool NimBLEClient::connect(const NimBLEAddress &address, uint8_t type, bool refr
 
     int rc = 0;
     m_peerAddress = address;
-    m_peerType = type;
+    m_peerAddrType = type;
 
     ble_addr_t peerAddrt;
     memcpy(&peerAddrt.val, address.getNative(),6);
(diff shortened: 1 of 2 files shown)


h2zero:
Can we change the naming to peerAddrType? It's a bit more typing but much clearer as to what the type means. 

The variable would then be m_peerAddrType and the method getPeerAddrType().

wakwak-koba:
I don't care about this, so you can do what you want !

h2zero:
Thank you for the update, looks good!

> I don't care about this, so you can do what you want !

Not sure of your meaning here, seems as though you're angry? Maybe a translation issue?

wakwak-koba:
Maybe I should have used 'mind' instead of 'care' ...
I'm sorry in bad English

h2zero:
@wakwak-koba No worries, I thought that may have been the case. Just wanted to check that I didn't upset you somehow.

As for this PR I have created a different one #104 that I think will be more generic. As well I have created a branch, feature/client-set-address to accompany it. These together should provide the functionality you requested in #98.

wakwak-koba:
I saw #104, I think it's very cool !
Then, isn't it necessary to review the arguments of BLEClient::connect()?

`
    bool                                        connect(const NimBLEAddress &address, uint8_t type = BLE_ADDR_PUBLIC,
                                                        bool refreshServices = true);
`

Since it contains `type` in `NimBLEAddress`, I think it will cause confusion if it exists in the argument of `connect`.
Should `type` be excluded from arguments?

`
    bool                                        connect(const NimBLEAddress &address, bool refreshServices = true);
`

h2zero:
The type has been removed from connect in the PR #106 which also incorporates the other parts of your request.

h2zero:
I merged the other one, is there any need to keep this open?

wakwak-koba:
This issue has been resolved in PR #106 and PR #104

PR #104, opened 2020-07-30
Store address type in NimBLEAddress.

h2zero:
Currently only the address value is stored in NimBLEAddress.

This adds the address type as a member variable and adds a type parameter to the constructor.

New NimBLEAddress::getType() method added to access the type value.

### DIFF
--- a/src/NimBLEAddress.cpp
+++ b/src/NimBLEAddress.cpp
@@ -33,6 +33,7 @@ static const char* LOG_TAG = "NimBLEAddress";
  */
 NimBLEAddress::NimBLEAddress(ble_addr_t address) {
     memcpy(m_address, address.val, 6);
+    m_addrType = address.type;
 } // NimBLEAddress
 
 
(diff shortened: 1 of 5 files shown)


1technophile:
Hello,

Is there a way to remove the type from the address,  Address: 2c:15:fb:d9:ca:dc **type: 1**, currently I'm getting:
```
Devices found: 9
Scan done!
Address: 2c:15:fb:d9:ca:dc type: 1 
Advertised Device: Name: , Address: 2c:15:fb:d9:ca:dc type: 1, manufacturer data: 06000109200275c149f28da0942a67f4378ea59e17c312e6851165a1d2
advType:  
Address: 60:59:f1:b6:2c:6e type: 1 
Advertised Device: Name: , Address: 60:59:f1:b6:2c:6e type: 1, manufacturer data: e0000106ca56fe8f, serviceUUID: 0xfe9f
Service Data:
UUID: 0xfe9f, Data:  
Address: a4:c1:38:d5:d4:98 type: 0 
Advertised Device: Name: LYWSD03MMC, Address: a4:c1:38:d5:d4:98 type: 0
```
And as I'm using the id as an unique device identifier it breaks my implementation and I need to remove the type from the string.
I would prefer to have directly the address alone and add the type by the getType method when needed.

h2zero:
Hi @1technophile, I will revert that change from this PR as it was only added for logging purposes. Although useful info, it's obtainable other ways if needed and I did not consider that someone could be using the string for their data.

#109

1technophile:
Perfect, thanks!

PR #106, opened 2020-07-30
Client: Add option to manually set peer address

h2zero:


wakwak-koba:
I'm pulling and testing this branch and my sample sketch worked very well.
https://github.com/wakwak-koba/NimBLE-Arduino/blob/master/examples/NimBLE_CatchNotifiesReConnect/NimBLE_CatchNotifiesReConnect.ino

Before this, the version of `BLEClient::connect(false)` you told me didn't work properly on my BLE device, so I'm refreshing it every time I reconnect.

h2zero:
Thanks for testing, I will merge it soon if you're happy with it. 

>Before this, the version of BLEClient::connect(false) you told me didn't work properly on my BLE device, so I'm refreshing it every time I reconnect.

You're example is making the first connection in loop, which will obviously need to get the services the first time, but is wasteful of the devices battery afterward. What you could do is use `pClient->connect(false)` and then test if you have the services already before calling getServices(true). 

```
auto pServices = pClient->getServices();
if(pServices->size() == 0) {
    pClient->getServices(true);
}
```
Same thing for characteristics.

wakwak-koba:
The method you taught me the other day (https://github.com/h2zero/NimBLE-Arduino/files/4977993/catchNotify_reconnect.ino.txt) is successful until reconnect and resist, but the actual Notify is It was in a state of not flying.
I'm sure it's because of the BLE device I use for testing.

h2zero:
Interesting, only thing I can think of, without seeing a log of the service handles to prove it, is the device you’re connecting to changes it’s service handles. If that’s the case, what you’re doing would be the correct action. 

Do they disconnect often?

wakwak-koba:
When it remains idle for about 30 seconds, it automatically disconnects and enters sleep mode to save power.

h2zero:
After looking at the instructions it seems to pair after connecting. You might consider enabling bonding in your setup to make reconnecting better, also may explain why you need to get the services again. 

I think I have one of these somewhere that I took apart, got it from the dollar store for $1, If I can find it I'll play around.

Other than that, is this PR enough for your needs?

wakwak-koba:
Yes, that's enough for me !

Issue #116, opened 2020-08-17
Bond pairing again fails with BLE_ERR_PINKEY_MISSING.


gkoh:
Firstly, thanks for a lovely library, really nice to use.

I ran into a problem, which took a while to track down.
If I successfully bond pair once, then try again, 'secureConnection()' fails.
I hacked in to see the taskData return code and it's BLE_ERR_PINKEY_MISSING.

After searching up and down the stack, I finally worked out it's because I never deleted the ESP32's bonding keys before re-attempting the pairing.

Once I added the following to my client removal code:
```
  {
    ble_addr_t peerAddr;
    memcpy(&peerAddr.val, m_Address.getNative(), 6);
    peerAddr.type = m_Address.getType();
    ble_gap_unpair(&peerAddr);
  }
```

it started working as expected.

To abstract this from the user, we could add a function to unpair a NimBLEAddress.
Additionally we could try to unpair automatically if secureConnection() fails, we're in the descriptor/characteristic retry loop and the error is 'PINKEY_MISSING' or 'REPEAT_PAIRING_RETRY' similar.

It's unclear to me how the REPEAT_PAIRING_RETRY error from the depths of the Nimble stack turns into PINKEY_MISSING.

h2zero:
Hi @gkoh, thanks for the detailed description. When you say "REPEAT_PAIRING_RETRY" are you referring to  BLE_GAP_EVENT_REPEAT_PAIRING? If that is the case, we can simply handle that as is done with the server code.

 The client does not have a handler for that event yet (totally forgot about that, thanks for the reminder).

>It's unclear to me how the REPEAT_PAIRING_RETRY error from the depths of the Nimble stack turns into PINKEY_MISSING.

What happens (I think), is when you're already paired and trying to re-pair it triggers the REPEAT_PAIRING event, which we are not handling at the moment. So without the handler to delete the previous pairing info we end up not sending any key/pin info, thus the error.

gkoh:
> Hi @gkoh, thanks for the detailed description. When you say "REPEAT_PAIRING_RETRY" are you referring to BLE_GAP_EVENT_REPEAT_PAIRING? If that is the case, we can simply handle that as is done with the server code.

I recall I traced it to an rc of BLE_GAP_REPEAT_PAIRING_RETRY somewhere, which I guess becomes the equivalent event.

> The client does not have a handler for that event yet (totally forgot about that, thanks for the reminder).

[...]

PR #117, opened 2020-08-20
Delete peer keys on key missing and retry.

gkoh:
Address #116.
On BLE_ERR_PINKEY_MISSING, delete the peer keys from the store and
retry the connection.


gkoh:
Tested against the following scenario:
- scan, then securely connect
- delete pairing from remote target
- scan, then securely connect again

Works seamlessly and transparently now, yay.

h2zero:
Thanks for this. Nice work!

gkoh:
> Thanks for this. Nice work!

No problems! Thanks for the guidance!

Issue #122, opened 2020-09-01
Use std::function in typedef of callback functions to enable callback to member functions


corneliusmunz:
I am using your library to connect to multiple BLE devices (LEGO hubs) at the same time. With your support we have managed to connect up to 9 devices. 

I have now another issue when connecting to multiple remote characteristics of several devices. **With the current implementation of the callback function i can only use a static function and not a member function as a callback**. This leads to the problem, that every connected device use the same static callback function and i can not preserve state or handle the notifications of different devices in a member function of a class. 

This problem with callbacks is described in several stackoverflow posts:
https://stackoverflow.com/questions/400257/how-can-i-pass-a-class-member-function-as-a-callback#402385
https://stackoverflow.com/questions/14189440/c-callback-using-class-member/14189561#14189561
https://cognitivewaves.wordpress.com/callbacks/

**One solution is to use the std::bind and std::function functionality which was introduced in C++11**

I have tested the following changes in the NimBLERemoteCharacteristic.h file:

1. Add` #include <functional>` to be able to use the std::function functionallity
2. Change the typedef of the callback as follows
`typedef std::function<void (NimBLERemoteCharacteristic* pBLERemoteCharacteristic,
                                uint8_t* pData, size_t length, bool isNotify)> notify_callback;`

With that changes i could use the "old" way of defining the callback function without any error
`_pRemoteCharacteristic->registerForNotify(notifyCallback); // nofifyCallback is a pointer to a static function
`
and i can use now with the `std::bind` functionality the "new" way of defining the callback function which enables me to use a member function instead of a static function
`_pRemoteCharacteristic->registerForNotify(std::bind(&Lpf2Hub::notifyCallback, this, _1, _2, _3, _4)); // notifyCallback is now a member function`

Maybe you can check this improvement and add it to one of your next releases. I think its a "transparent" change where the "old" static callback functions will work without changes in the client code which uses your library. I can also state a pull request with my changes that you can check it directly. I have added the follwing PR with the working changes (https://github.com/h2zero/NimBLE-Arduino/pull/123)

I know that the `registerForNotify` method is marked as deprecated an i should replace it with the `subscribe` method but in principal this callback thing is also valid for the new `subscribe` method

@giancann, @lucadentella: With this issue solved, the legoino issue (https://github.com/corneliusmunz/legoino/issues/19) could also be solved but it is a prerequisite to solve it in the NimBLE library

@h2zero : Thanks in advance to have a look on that issue!

[...]

PR #123, opened 2020-09-01
Change typedef of callback to enable the usage of member function as callback functions 

corneliusmunz:
With this change the callback function could be also a member function and not only a static function. This is needed if you connect to several devices and want to preserve the state in your objects. The detailed description of the issue could be found here: https://github.com/h2zero/NimBLE-Arduino/issues/122


h2zero:
Thanks for this! 

Looks good to me, verified functional without API changes.

PR #136, opened 2020-10-09
Fixed incomplete payload size calculation

### LINKED ISSUE #135: What should I do if the advertising size exceeds 31 bytes ?
If I use 128-bit service class UUIDs, the payload size will be large and there will be little room to store other items.
Now when the payload exceeds 31 bytes, it will abort as follows:
```
CRIT NimBLEAdvertising: "error setting advertisement data; rc=4, "
```
If I use uuid128, specify Appearance, the local name can only use up to 4 characters.
What should I do?

wakwak-koba:
I tried to fix #135 in my own way.

### DIFF
--- a/src/NimBLEAdvertising.cpp
+++ b/src/NimBLEAdvertising.cpp
@@ -41,8 +41,8 @@ NimBLEAdvertising::NimBLEAdvertising() {
     m_advData.name                   = (uint8_t *)name;
     m_advData.name_len               = strlen(name);
     m_advData.name_is_complete       = 1;
-    m_scanData.tx_pwr_lvl_is_present = 1;
-    m_scanData.tx_pwr_lvl            = NimBLEDevice::getPower();
+    m_advData.tx_pwr_lvl_is_present = 1;
+    m_advData.tx_pwr_lvl            = NimBLEDevice::getPower();
     m_advData.flags                  = (BLE_HS_ADV_F_DISC_GEN | BLE_HS_ADV_F_BREDR_UNSUP);
     m_advData.appearance             = 0;
     m_advData.appearance_is_present  = 0;
@@ -264,12 +264,16 @@ void NimBLEAdvertising::start(uint32_t duration, void (*advCompleteCB)(NimBLEAdv
 
     if (!m_customAdvData && !m_advDataSet) {
         //start with 3 bytes for the flags data
-        uint8_t payloadLen = 3;
+        uint8_t payloadLen = (2 + 1);
+        if(m_advData.appearance_is_present)
+            payloadLen += (2 + BLE_HS_ADV_APPEARANCE_LEN);
+        if(m_advData.tx_pwr_lvl_is_present)
+            payloadLen += (2 + 1);
 
         for(auto &it : m_serviceUUIDs) {
             if(it.getNative()->u.type == BLE_UUID_TYPE_16) {
                 int add = (m_advData.num_uuids16 > 0) ? 2 : 4;
-                if((payloadLen + add) > 31){
+                if((payloadLen + add) > BLE_HS_ADV_MAX_SZ){
                     m_advData.uuids16_is_complete = 0;
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


h2zero:
This looks good, my only concern is that prioritizing tx_pwr_lvl over the name field will limit the name length by another 3 bytes if the app isn't using scan response. Some folks may have an issue with that.

wakwak-koba:
I didn't really think about the case where RSP wasn't possible, but I also understand the desire to reserve more space for device names, even with just 3 bytes.
So I modified the code to disable tx_pwr_lvl when RSP is not possible and the device name length is exceeded.

Please check again !

h2zero:
Just one other thing we should touch while here. We need to ensure the name does not exceed 29 bytes when moved to scan response. For that we should set the len byte to 29 and mark it as incomplete. That's what some of the removed code did, other than that it looks good.

> I also understand the desire to reserve more space for device names, even with just 3 bytes

Yeah, 3 bytes doesn't seem like much but it's 10% of the advertising space, thanks for changing it.

wakwak-koba:
I completely forgot the existence of device names larger than 29 bytes.
sorry...

It may be more beautiful to use `min(a, b)`, but I will add it using if for easy understanding.

src/NimBLEAdvertising.cpp
```diff
@@ -333,48 +337,37 @@ void NimBLEAdvertising::start(uint32_t duration, void (*advCompleteCB)(NimBLEAdv
         }
 
         // check if there is room for the name, if not put it in scan data
-        if((payloadLen + m_advData.name_len) > 29) {
+        if((payloadLen + (2 + m_advData.name_len)) > BLE_HS_ADV_MAX_SZ) {
             if(m_scanResp){
                 m_scanData.name = m_advData.name;
                 m_scanData.name_len = m_advData.name_len;
+                if(m_scanData.name_len > BLE_HS_ADV_MAX_SZ - 2) {
+                    m_scanData.name_len = BLE_HS_ADV_MAX_SZ - 2;
```
h2zero, line 345: We need to add `m_scanData.name_is_complete = 0;` here. Since we are cutting off the name it should be marked incomplete.

src/NimBLEAdvertising.cpp
```diff
@@ -333,48 +337,37 @@ void NimBLEAdvertising::start(uint32_t duration, void (*advCompleteCB)(NimBLEAdv
...
+                }
                 m_scanData.name_is_complete = m_advData.name_is_complete;
```
h2zero: To accommodate the above comment this should be inside an `else` block and =1.

src/NimBLEAdvertising.cpp
```diff
@@ -333,48 +337,37 @@ void NimBLEAdvertising::start(uint32_t duration, void (*advCompleteCB)(NimBLEAdv
...
+                if(m_advData.tx_pwr_lvl_is_present) {
+                    m_advData.tx_pwr_lvl = 0;
+                    m_advData.tx_pwr_lvl_is_present = 0;
+                    payloadLen -= (2 + 1);
+                }
                 // if not using scan response just cut the name down
                 // leaving 2 bytes for the data specifier.
-                m_advData.name_len = (29 - payloadLen);
+                if(m_advData.name_len > (BLE_HS_ADV_MAX_SZ - payloadLen - 2)) {
+                    m_advData.name_len = (BLE_HS_ADV_MAX_SZ - payloadLen - 2);
```
h2zero, line 362: I think we can move the `m_advData.name_is_complete = 0;` from below to here.

src/NimBLEAdvertising.cpp
```diff
@@ -333,48 +337,37 @@ void NimBLEAdvertising::start(uint32_t duration, void (*advCompleteCB)(NimBLEAdv
...
+                }
             }
             m_advData.name_is_complete = 0;
```
h2zero: If moved as per above, this can be removed.

h2zero (changes requested):
Almost there, just need to do a couple minor adjustments. Please have a look at the comments I left and let me know if the changes work for you.

wakwak-koba:
There was insufficient consideration for `name_is_complete`.
I'm sure this is ok...

src/NimBLEAdvertising.cpp
wakwak-koba: I'm sorry I'm incomplete.
I just fixed it

If you find another bug, you can fix it.

h2zero (approved):

h2zero:
Sorry for the delay, been away for a while dealing with other things. 

Thanks for fixing that up, I will merge now.

PR #137, opened 2020-10-14
Added possibility to define custom stack size

piaskowyk:


h2zero:
Why are these changes needed for this? Does it not take effect if `CONFIG_BT_NIMBLE_TASK_STACK_SIZE` is changed in `nimconfig.h`?

piaskowyk:
I needed an increased stack size because action performed on BLE threads needed more memory than was available.
I found another solution so we can close this PR.

PR #152, opened 2020-11-30
Update NimBLERemoteCharacteristic.cpp

deepak4395:
BLE itags don't return descriptor values so bypassing them to register notify callback function.

### DIFF
--- a/src/NimBLERemoteCharacteristic.cpp
+++ b/src/NimBLERemoteCharacteristic.cpp
@@ -510,18 +510,19 @@ int NimBLERemoteCharacteristic::onReadCB(uint16_t conn_handle,
  */
 bool NimBLERemoteCharacteristic::setNotify(uint16_t val, notify_callback notifyCallback, bool response) {
     NIMBLE_LOGD(LOG_TAG, ">> setNotify(): %s, %02x", toString().c_str(), val);
-
+/*
     NimBLERemoteDescriptor* desc = getDescriptor(NimBLEUUID((uint16_t)0x2902));
     if(desc == nullptr) {
         NIMBLE_LOGE(LOG_TAG, "<< setNotify(): Could not get descriptor");
         return false;
     }
-
+*/
     m_notifyCallback = notifyCallback;
 
     NIMBLE_LOGD(LOG_TAG, "<< setNotify()");
 
-    return desc->writeValue((uint8_t *)&val, 2, response);
+  //  return desc->writeValue((uint8_t *)&val, 2, response);
+	return true;
 } // setNotify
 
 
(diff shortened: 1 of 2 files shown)


chegewara:
This may work in your case, but not the other cases. You cant change library like that just for 1 case.

h2zero:
Thanks for this but as @chegewara said this may cause problems with other uses. 

However a change could be made to support this, I'll do some testing and get back to you.

src/nimconfig.h
```diff
@@ -102,7 +102,7 @@
 #define CONFIG_BT_NIMBLE_MEM_ALLOC_MODE_INTERNAL 1
 
 /** @brief Sets the number of simultaneous connections (esp controller max is 9) */
-#define CONFIG_BT_NIMBLE_MAX_CONNECTIONS 3
+#define CONFIG_BT_NIMBLE_MAX_CONNECTIONS 8
```
h2zero, line 105: This change won't be merged as it consumes resources that may not be needed by other users. Better to leave this at the default value and users can change this for their own needs.

src/NimBLERemoteCharacteristic.cpp
```diff
@@ -510,18 +510,19 @@ int NimBLERemoteCharacteristic::onReadCB(uint16_t conn_handle,
...
     NIMBLE_LOGD(LOG_TAG, ">> setNotify(): %s, %02x", toString().c_str(), val);
-
+/*
     NimBLERemoteDescriptor* desc = getDescriptor(NimBLEUUID((uint16_t)0x2902));
     if(desc == nullptr) {
         NIMBLE_LOGE(LOG_TAG, "<< setNotify(): Could not get descriptor");
         return false;
     }
-
+*/
```
h2zero, line 519: Instead of removing this block we should move it to after the callback is set on the next line.

This should accommodate your use case as well as others.

deepak4395:
> This may work in your case, but not the other cases. You cant change library like that just for 1 case.

Can you please suggest a better way to do it?

deepak4395:
> Thanks for this but as @chegewara said this may cause problems with other uses.
> 
> However a change could be made to support this, I'll do some testing and get back to you.
ohk

h2zero:
Sorry, new github code review has an extra step now that I didn't notice, made comments days ago that didn't get posted.

src/NimBLERemoteCharacteristic.cpp
deepak4395: Ohk

src/nimconfig.h
deepak4395: Ohk

h2zero:
Not sure why you closed this? If you adjust the code as I suggested and let me know if it works for you I would merge it.

h2zero:
@deepak4395 Since this was closed, I have pushed a new commit 8310b7b that will accommodate your use and credited you as the author. I hope this suits your needs. Please open another PR if it is insufficient.

PR #169, opened 2021-01-08
Add examples of secured and authenticated connection

mblasee:


h2zero:
Looks good to me, thank you!

I will merge this soon. I am pushing some bug fixes and updating examples as well, I'd like to merge after that is done so I don't have to deal with conflicts :).

selman-nus:
NimBLE_Secure_Client.ino
`NimBLERemoteCharacteristic *pNonSecureCharacteristic = pService->getCharacteristic("1234");
NimBLERemoteCharacteristic *pSecureCharacteristic         = pService->getCharacteristic("1235");
`

Hi, seems like there is no difference between the secure and nonsecure characteristics. Am i missing something?

mblasee:
If you watch the server you will see that the 1235 characteristic has NIMBLE_PROPERTY::READ_ENC and NIMBLE_PROPERTY::READ_AUTHE flags. That means if you change the passkey either in the client or server you wont be able to read the 1235 characteristic, only the 1234 one as that one is not secured.

selman-nus:
oh alright, I thought it was a generic description, I didn't get the server and the client examples were designed in pairs. Thanks for the clarification!

PR #171, opened 2021-01-09
Find characteristic endhandle using the next characterisic handle. (#170)

### LINKED ISSUE #170: Some devices fail setNotify
First of all, there are items to be confirmed before the main subject.

https://github.com/h2zero/NimBLE-Arduino/blob/bc52eac5618c45bcfc00b1004bc25b3b1709f0ae/src/NimBLERemoteCharacteristic.cpp#L529-L533

If descriptor 0x2902 does not exist, `setNotify()` returns `true`, which is correct, isn't it `false`?
That's because I found that some devices couldn't get descriptor 0x2902 even though `canNotify() == true`.

I don't think it's similar to the 16bit uuid problem I had before, but do you have any idea?

The device used for evaluation is [this](https://www.aliexpress.com/item/32758382392.html).

wakwak-koba:
fixed #170.

The handle termination of RemoteCharacteristic is now calculated and used instead of the termination of service.

### DIFF
--- a/src/NimBLERemoteCharacteristic.cpp
+++ b/src/NimBLERemoteCharacteristic.cpp
@@ -232,7 +232,7 @@ bool NimBLERemoteCharacteristic::retrieveDescriptors(const NimBLEUUID *uuid_filt
     ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
 
     if(taskData.rc != 0) {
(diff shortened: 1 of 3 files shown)


h2zero:
Thanks for this! I will merge this soon. I'm working on some bug fixes at the moment and hope to complete them before merging PR's.

Issue #172, opened 2021-01-09
Could you give me some advice to use xbox controller: HID client + L2CAP + Privacy


asukiaaa:
I try to get HID notification from xbox controller (for series x, type 1914) but not succeeded.
I can read HID info map and register notification (I checked CCCD become 0100 from 0000) but any notification is not published and a led on the controller is blinking (that means not finish communication setting).

[A driver for linux (xpadneo)](https://github.com/atar-axis/xpadneo) requires [privacy settings](https://github.com/atar-axis/xpadneo/issues/166#issuecomment-747010866)(setting `Privacy=device` to `etc/bluetooth/main.conf`) and seems using [L2CAP](https://github.com/atar-axis/xpadneo/blob/d55e6d42ecb53f3ebe91e7a43574c35e79146dfd/misc/kernel_patches/0001-fix_bluetooth_reconnect.patch#L24) so I think they are needed to communicate with the controller.

Questions

- Does this library support L2CAP?

I found [code about L2CAP on this project](https://github.com/h2zero/NimBLE-Arduino/blob/master/src/nimble/host/mesh/src/src/ble_l2cap_priv.h) but I'm not sure it is activated on client mode.

- Is there any option to use privacy setting on client mode?

I found privacy setting on this project.
However if I call `ble_hs_pvcy_rpa_config(NIMBLE_HOST_ENABLE_RPA);` (or `NIMBLE_HOST_ENABLE_NRPA`) after `NimBLEDevice::init("");` with setting `BLE_HOST_BASED_PRIVACY=1` as a build flag, esp32 cannot read information from characteristics after connection with the controller.

If you have some idea to start notification of the controller, I'm happy if you share that.

Code that I tried: https://github.com/asukiaaa/esp32-client-for-xbox-controller-with-nim-ble/blob/master/src/main.cpp

Thank you.

h2zero:
Hi @asukiaaa there should not be a need to use L2CAP directly as NimBLE should take care of that when needed.

If you need to use RPA [here is the details](https://github.com/h2zero/NimBLE-Arduino/blob/537629f471eea1017b28a2482c64a0e7cfbeed55/src/host/ble_hs_pvcy.h#L37), as the API has not be written for this yet, it will require you to make a few modifications to `NimBLEDevice`.

asukiaaa:
Thank you very much for the information.
I succeeded in receiving notifications from the controller!

I could set values with using the following code except for `m_own_addr_type`.
[...]

PR #174, opened 2021-01-09
Add option to use resolvable and non-resolvable private address.

asukiaaa:
To use RPA or NRPA without editing code of library.
Reference: https://github.com/h2zero/NimBLE-Arduino/issues/172

### DIFF
--- a/src/NimBLEDevice.cpp
+++ b/src/NimBLEDevice.cpp
@@ -718,19 +718,17 @@ void NimBLEDevice::setSecurityCallbacks(NimBLESecurityCallbacks* callbacks) {
  */
 void NimBLEDevice::setOwnAddrType(uint8_t own_addr_type, bool useNRPA) {
     m_own_addr_type = own_addr_type;
-    if (own_addr_type != BLE_OWN_ADDR_PUBLIC) {
-        setSecurityInitKey(BLE_SM_PAIR_KEY_DIST_ID);
-        setSecurityRespKey(BLE_SM_PAIR_KEY_DIST_ID);
-    }
     switch (own_addr_type) {
         case BLE_OWN_ADDR_PUBLIC:
             ble_hs_pvcy_rpa_config(NIMBLE_HOST_DISABLE_PRIVACY);
             break;
         case BLE_OWN_ADDR_RANDOM:
+            setSecurityInitKey(BLE_SM_PAIR_KEY_DIST_ENC | BLE_SM_PAIR_KEY_DIST_ID);
             ble_hs_pvcy_rpa_config(useNRPA ? NIMBLE_HOST_ENABLE_NRPA : NIMBLE_HOST_ENABLE_RPA);
             break;
         case BLE_OWN_ADDR_RPA_PUBLIC_DEFAULT:
         case BLE_OWN_ADDR_RPA_RANDOM_DEFAULT:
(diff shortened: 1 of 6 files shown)


h2zero:
This is a good beginning, however like I said in your issue thread this API should be in the NimBLEDevice class and all the other places this value is used should call a method on it to get the value.

Here are the other places:
https://github.com/h2zero/NimBLE-Arduino/blob/bc52eac5618c45bcfc00b1004bc25b3b1709f0ae/src/NimBLEClient.cpp#L197
https://github.com/h2zero/NimBLE-Arduino/blob/5cb454bfce3bf753e26d00817b7bf66cc0de447a/src/NimBLEAdvertising.cpp#L476

Also, in the call to set the address type we should make the appropriate call to `ble_hs_pvcy_rpa_config();` Not a whole lot to change and I would be more than happy to merge it with those changes.

asukiaaa:
@h2zero 
Thank you for showing the specific code.
Maybe I got it.
I update code to handle m_own_addr through NimBLEDevice.
How about this?

src/NimBLEDevice.cpp
```diff
@@ -705,6 +707,36 @@ void NimBLEDevice::setSecurityCallbacks(NimBLESecurityCallbacks* callbacks) {
...
+ * The available bits are defined as:
+ * * 0x00: BLE_OWN_ADDR_PUBLIC
+ * * 0x01: BLE_OWN_ADDR_RANDOM
+ * * 0x02: BLE_OWN_ADDR_RPA_PUBLIC_DEFAULT
+ * * 0x03: BLE_OWN_ADDR_RPA_RANDOM_DEFAULT
+ */
+void NimBLEDevice::setOwnAddrType(uint8_t own_addr_type, bool useNRPA) {
+    m_own_addr_type = own_addr_type;
+    if (own_addr_type != BLE_OWN_ADDR_PUBLIC) {
+        setSecurityInitKey(BLE_SM_PAIR_KEY_DIST_ID);
```
h2zero: This needs to be `setSecurityInitKey(BLE_SM_PAIR_KEY_DIST_ENC | BLE_SM_PAIR_KEY_DIST_ID);` or we won't be able to encrypt connections.

src/NimBLEDevice.cpp
```diff
@@ -705,6 +707,36 @@ void NimBLEDevice::setSecurityCallbacks(NimBLESecurityCallbacks* callbacks) {
...
+        setSecurityRespKey(BLE_SM_PAIR_KEY_DIST_ID);
```
h2zero: This line can be removed, as the default setting in `init()` is ` BLE_SM_PAIR_KEY_DIST_ENC | BLE_SM_PAIR_KEY_DIST_ID`

h2zero:
Looks great! thank you for this. just a couple minor comments to address.

asukiaaa:
Thank you for checking.
I call `setSecurityInitKey(BLE_SM_PAIR_KEY_DIST_ENC | BLE_SM_PAIR_KEY_DIST_ID);` in `setOwnAddrType`.

h2zero:
Looks good to me, thank you!

h2zero:
Just noticed this before merging:

https://github.com/h2zero/NimBLE-Arduino/blob/5cb454bfce3bf753e26d00817b7bf66cc0de447a/src/NimBLEAdvertising.cpp#L481

That line needs to have the address type change as well, my mistake, it should have been mentioned earlier.

asukiaaa:
Thank you for checking carefully.
I updated.

PR #178, opened 2021-01-16
silence compiler error for espsolo1 arduino framework 

btsimonh:
where ipc does not exist.

would be great to drop this in so Tasmota does not have to use a modified version....

(@jason8266 - pls comment if necessary)


h2zero:
Thanks for this. Makes sense that the IPC header would not be available for the solo, which should be supported here.

Will merge this shortly.

Jason2866:
Thx! Great :-)

Issue #128, opened 2020-09-15
Incomplete payload in NimBLEAdvertisedDevice


corneliusmunz:
During the Scan process of my library  i detect that the behaviour of the payload in the NimBLE lib is different than in the former BLE lib. The payload contains only the last part of the overall payload and it seems to be that the payload is overwritten. The address to the pointer remains the same but the content and content length differs. 

payload of "old" non NimBLE based lib:
```
2-1-6-11-7-23-D1-BC-EA-5F-78-23-16-DE-EF-12-12-23-16-0-0-9-FF-97-3-0-41-7-80-63-0-5-12-10-0-20-0-2-A-0-B-9-6D-79-54-72-61-69-6E-48-75
```

payload of NimBLE-Arduino lib:
```
5-12-10-0-20-0-2-A-0-B-9-6D-79-54-72-61-69-6E-48-75
```

The payload is fetched after the callback is executed with the ```NimBLEAdvertisedDevice``` ```getPayload()```function

I have enabled the log files and added some log messages (payload address and payload length) with the following output
```
13:43:34.066 -> I NimBLEScan: "NEW DEVICE FOUND: 90:84:2b:03:19:7f"
13:43:34.066 -> D NimBLEAdvertisedDevice: "payload address: 3ffc86ac"
13:43:34.066 -> D NimBLEAdvertisedDevice: "payload length: 31"
13:43:34.066 -> D NimBLEAdvertisedDevice: "payload: 020106110723d1bcea5f782316deef12122316000009ff9703004107806300"
13:43:34.066 -> I NimBLEScan: "UPDATING PREVIOUSLY FOUND DEVICE: 90:84:2b:03:19:7f"
13:43:34.066 -> D NimBLEAdvertisedDevice: "payload address: 3ffc86ac"
13:43:34.102 -> D NimBLEAdvertisedDevice: "payload length: 21"
13:43:34.102 -> D NimBLEAdvertisedDevice: "payload: 051210002000020a000b096d79547261696e487562"
```
When looking at the address and the payload length (additional log messages added in the NimBLE lib) it seems that the original location is overwritten. I seems that the ```parseAdvertisement``` function is called twice before the callback function is executed. 

[...]

PR #181, opened 2021-01-25
Refactor advertised device

### LINKED ISSUE #165: NimBLEScan : NimBLEAdvertisedDevice getpayload() returns incomplete packet
Using an ESP32 and the example scan code, I have noticed a difference in the payload length compared with the original nkolban library. I used the example scan code from both this library and the nkolban one and for one device I got a different length payload. It appears that part of the payload is missing or incomplete as I have compared the expected data with the manufacturer's data sheet and the payload seems to have skipped about 30 bytes. 

From the snippet below - the payload starts at the two 16s onwards, whereas the full payload is expected:
image

I have another type of sensor which is decoded the same for both this library and the nkolban one, so maybe some logic there for iBeacon types, which the first sensor is of this type. This issue fortunately is easy to deal with as the actual data is in the portion of the payload that is retrieved.

h2zero:
This PR is a complete refactoring of the advertisedDevice class as used in scan results. Resolves #165 and #128.

The purpose is to provide access to all the advertised data through the getPayload method without increasing the heap consumed. In fact, this reduces the consumption by approximately 25% per device while providing full payload data. 

Highlights:

* Stores complete advertisement payload and only performs parsing on demand. The advertised data is no longer parsed automatically as it is discovered, instead it will be parsed to find only the data requested by the user application when it makes a call to do so. This saves processing time and approximately 2.4k in flash size, with the new methods listed below included.

Add more advertised device field methods:
  
* Adds haveAdvInterval/getAdvInterval - checks if the interval is advertised / gets the advertisement interval value.
    
* Adds haveConnParams/getMinInterval/getMaxInterval - checks if the parameters are advertised / get min value / get max value.
    
* Adds haveURI/getURI - checks if a URI is advertised / gets the URI data.
    
* Adds haveTargetAddress/getTargetAddressCount/getTargetAddress(index) - checks if a target address is present / gets a count of the addresses targeted / gets the address of the target at index.

**Requires more testing, please provide any feedback you may have.**

### DIFF
--- a/src/NimBLEAdvertisedDevice.cpp
+++ b/src/NimBLEAdvertisedDevice.cpp
@@ -28,25 +28,14 @@ static const char* LOG_TAG = "NimBLEAdvertisedDevice";
 /**
  * @brief Constructor
  */
-NimBLEAdvertisedDevice::NimBLEAdvertisedDevice() {
+NimBLEAdvertisedDevice::NimBLEAdvertisedDevice() :
+    m_payload(62,0)
+{
     m_advType          = 0;
-    m_appearance       = 0;
-    m_manufacturerData = "";
-    m_name             = "";
     m_rssi             = -9999;
-    m_txPower          = 0;
-    m_payloadLength    = 0;
-    m_payload          = nullptr;
-
-    m_haveAppearance       = false;
-    m_haveManufacturerData = false;
-    m_haveName             = false;
-    m_haveRSSI             = false;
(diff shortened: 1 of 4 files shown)


doudar:
Wow! This looks awesome!
I've got a little more time than normal this week (Wife is on vacation! :) ) So maybe I'll try to adapt my code and do some testing.

kind3r:
So far, in my use case, all is well.

PR #182, opened 2021-01-29
[NimBLEHIDDevice] Modern Apple devices require `READ_ENC` and `WRITE_ENC`

wakwak-koba:
I noticed while evaluating the [ESP32-NimBLE-Keyboard](https://github.com/wakwak-koba/ESP32-NimBLE-Keyboard) that I ported.
I'm having trouble pairing from iOS, or I can't receive data even if pairing is successful.

As a result of various attempts, the end result is that modern Apple devices may refuse to communicate in clear text.

Specifically, in HID-Keyboard (ServiceUUID: 0x1812), it was found that READ_ENC must be specified in Report Input (CharacteristicUUID: 0x2a4d).
Furthermore, Apple devices did not receive data unless `*_ENC` was specified during `createDescriptor()`.

If adding `*_ENC` to the default value of `createDescriptor()` does not cause another problem, I think it would be more convenient to add it on the NimBLE-Arduino side, so instead of `NimHIDDevice.cpp`, Fixed the `NimBLECharacteristic.h` side.

Please let me know your opinions !

### DIFF
--- a/src/NimBLECharacteristic.h
+++ b/src/NimBLECharacteristic.h
@@ -61,13 +61,13 @@ class NimBLECharacteristic {
 public:
     NimBLEDescriptor* createDescriptor(const char* uuid,
                                        uint32_t properties =
-                                       NIMBLE_PROPERTY::READ |
-                                       NIMBLE_PROPERTY::WRITE,
+                                       NIMBLE_PROPERTY::READ | NIMBLE_PROPERTY::READ_ENC |
+                                       NIMBLE_PROPERTY::WRITE | NIMBLE_PROPERTY::WRITE_ENC,
                                        uint16_t max_len = 100);
     NimBLEDescriptor* createDescriptor(const NimBLEUUID &uuid,
                                        uint32_t properties =
-                                       NIMBLE_PROPERTY::READ |
-                                       NIMBLE_PROPERTY::WRITE,
+                                       NIMBLE_PROPERTY::READ | NIMBLE_PROPERTY::READ_ENC |
+                                       NIMBLE_PROPERTY::WRITE | NIMBLE_PROPERTY::WRITE_ENC,
                                        uint16_t max_len = 100);
 
     NimBLEDescriptor* getDescriptorByUUID(const char* uuid);
--- a/src/NimBLECharacteristic.h
+++ b/src/NimBLECharacteristic.h
@@ -61,13 +61,13 @@ class NimBLECharacteristic {
 public:
     NimBLEDescriptor* createDescriptor(const char* uuid,
                                        uint32_t properties =
-                                       NIMBLE_PROPERTY::READ | NIMBLE_PROPERTY::READ_ENC |
-                                       NIMBLE_PROPERTY::WRITE | NIMBLE_PROPERTY::WRITE_ENC,
+                                       NIMBLE_PROPERTY::READ |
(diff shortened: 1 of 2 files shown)


src/NimBLEHIDDevice.cpp
```diff
@@ -128,7 +128,7 @@ void NimBLEHIDDevice::hidInfo(uint8_t country, uint8_t flags) {
  * @return pointer to new input report characteristic
  */
 NimBLECharacteristic* NimBLEHIDDevice::inputReport(uint8_t reportID) {
-	NimBLECharacteristic* inputReportCharacteristic = m_hidService->createCharacteristic((uint16_t) 0x2a4d, NIMBLE_PROPERTY::READ | NIMBLE_PROPERTY::NOTIFY);
+	NimBLECharacteristic* inputReportCharacteristic = m_hidService->createCharacteristic((uint16_t) 0x2a4d, NIMBLE_PROPERTY::READ | NIMBLE_PROPERTY::NOTIFY | NIMBLE_PROPERTY::READ_ENC);
```
h2zero, line 131: This is fine, makes sense to do.

src/NimBLECharacteristic.h
```diff
@@ -61,13 +61,13 @@ class NimBLECharacteristic {
...
+                                       NIMBLE_PROPERTY::READ | NIMBLE_PROPERTY::READ_ENC |
+                                       NIMBLE_PROPERTY::WRITE | NIMBLE_PROPERTY::WRITE_ENC,
                                        uint16_t max_len = 100);
     NimBLEDescriptor* createDescriptor(const NimBLEUUID &uuid,
                                        uint32_t properties =
-                                       NIMBLE_PROPERTY::READ |
-                                       NIMBLE_PROPERTY::WRITE,
+                                       NIMBLE_PROPERTY::READ | NIMBLE_PROPERTY::READ_ENC |
+                                       NIMBLE_PROPERTY::WRITE | NIMBLE_PROPERTY::WRITE_ENC,
                                        uint16_t max_len = 100);
```
h2zero: Most users will probably not want this, better to leave it up to them to change it.

wakwak-koba:
I understood.
For the `createDescriptor()` attribute, `BLEHIDDevice.cpp` handled it !

src/NimBLEHIDDevice.cpp
```diff
@@ -128,8 +128,8 @@ void NimBLEHIDDevice::hidInfo(uint8_t country, uint8_t flags) {
  * @return pointer to new input report characteristic
  */
 NimBLECharacteristic* NimBLEHIDDevice::inputReport(uint8_t reportID) {
-	NimBLECharacteristic* inputReportCharacteristic = m_hidService->createCharacteristic((uint16_t) 0x2a4d, NIMBLE_PROPERTY::READ | NIMBLE_PROPERTY::NOTIFY);
-	NimBLEDescriptor* inputReportDescriptor = inputReportCharacteristic->createDescriptor((uint16_t) 0x2908);
```
h2zero, line 132: Admittedly I do not know HID devices well, (learned quite a bit since this was added, thank you) a question though, is the 0x2908 descriptor not needed for the input report? I'm assuming not but wanted to verify.

h2zero: Oops, too quick. Sorry I just noticed the changes below, please disregard.

PR #189, opened 2021-02-02
Update migration guide to include Characteristic constructor now being private

### LINKED ISSUE #186: Update to migration guide
I've come to migrate and noticed that the change to the way characteristics are created has changed more than the migration guide suggests.  Using Bluedroid I subclassed `BLECharacteristic` and called `service->addCharacteristic()`.  Could I suggest including a note in the migration guide to say that the constructor is now private and the only way to create one is now `createCharacteristic`?

j4m3s:
Closes #186


h2zero:
Looks great, thanks for this!

PR #241, opened 2021-05-18
Add method to get the callbacks from a characteristic.

ketan:
This is useul to be able to unit/integration test the behavior of a
characteristic.


h2zero:
Thanks, I will merge this shortly.

PR #243, opened 2021-05-19
Call onSubscribe() after adding connection to subscribedVector

Bascy:
This allows for sending BLE messages in the onSubscribe() handler


h2zero:
Looks good to me, thanks!

PR #249, opened 2021-05-24
2.0.0 WIP

h2zero:
Too many changes to type; TODO later.


h2zero:
Damn, after all that I still can't beat @doudar , so close yet so far.

PR #265, opened 2021-07-18
[Server] Add callback for MTU change.

h2zero:
onMTUChange callback added that is called when a connection MTU is updated.


ketan:
Thank you for implementing this! Been looking for this.

PR #283, opened 2021-09-10
[Server][Client] Add function to set data length.

Mr-Mime:
This allows to set the data packet length.
Either the Server or Client can perform this action.

I am not quite sure about the `tx_time` limits.
When `tx_time` is greater or equal to  0x0849 I am getting an error in the logs which says it is an invalid value:
`E NimBLEServer: "Set data length error: 530, Invalid HCI Command Parameters"`

### DIFF
--- a/src/NimBLEClient.cpp
+++ b/src/NimBLEClient.cpp
@@ -437,6 +437,20 @@ void NimBLEClient::updateConnParams(uint16_t minInterval, uint16_t maxInterval,
 
 
 /**
+ * @brief Request an update of the data packet length.
+ * * Can only be used after a connection has been established.
+ * @param [in] tx_octets The preferred number of payload octets to use (Range 0x001B-0x00FB).
+ * @param [in] tx_time The preferred number of microseconds to use when transmitting a single packet (Range 0x0148-0x4290).
+ */
+void NimBLEServer::setDataLen(uint16_t tx_octets, uint16_t tx_time) {
+    int rc = ble_gap_set_data_len(m_conn_id, tx_octets, tx_time);
+    if(rc != 0) {
+        NIMBLE_LOGE(LOG_TAG, "Set data length error: %d, %s", rc, NimBLEUtils::returnCodeToString(rc));
+    }
+}// setDataLen
+
+
+/**
  * @brief Get detailed information about the current peer connection.
  */
 NimBLEConnInfo NimBLEClient::getConnInfo() {
--- a/src/NimBLEClient.cpp
+++ b/src/NimBLEClient.cpp
@@ -439,15 +439,19 @@ void NimBLEClient::updateConnParams(uint16_t minInterval, uint16_t maxInterval,
 /**
  * @brief Request an update of the data packet length.
  * * Can only be used after a connection has been established.
+ * @details Sends a data length update request to the server the client is connected to.
+ * The Data Length Extension (DLE) allows to increase the Data Channel Payload from 27 bytes to up to 251 bytes.
+ * The server needs to support the Bluetooth 4.2 specifications, to be capable of DLE.
  * @param [in] tx_octets The preferred number of payload octets to use (Range 0x001B-0x00FB).
- * @param [in] tx_time The preferred number of microseconds to use when transmitting a single packet (Range 0x0148-0x4290).
  */
-void NimBLEServer::setDataLen(uint16_t tx_octets, uint16_t tx_time) {
+void NimBLEClient::setDataLen(uint16_t tx_octets) {
+    uint16_t tx_time = (tx_octets + 14) * 8;
+
     int rc = ble_gap_set_data_len(m_conn_id, tx_octets, tx_time);
     if(rc != 0) {
         NIMBLE_LOGE(LOG_TAG, "Set data length error: %d, %s", rc, NimBLEUtils::returnCodeToString(rc));
     }
-}// setDataLen
+} // setDataLen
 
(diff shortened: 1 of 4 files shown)


h2zero:
Thanks, I'll test this out tonight. I'm curious if you have tested the speed difference and if you have any results to share?

The time limits have a restriction, I forget the value now but it is related to the maximum bit rate. I'll have to re-read up on it so I can add to the documentation.

Mr-Mime:
Oh yeah, totally forgot to note that.

I tested it with downloading a 100kB file.

Previously the download took 26-27 seconds.

With the packet size set to 251 (0x00fb), the download takes 14-15 seconds.
This equals an time reduction of 40-45%.

But I was not able to detect a difference, when changing the `tx_time`. 

This test was with an Huawei Y2, which support Bluetooth 4.2 Specification, which is needed for DLE.
However with an iPhone 6s, which also supports Bluetoth 4.2, I had no change in speed.
But iPhones are something for themself..., maybe one could find out more when sniffing the communication to see what the response of the iPhone is.

If I can do something to help testing, just let me know.

src/NimBLEClient.cpp
```diff
@@ -436,6 +436,20 @@ void NimBLEClient::updateConnParams(uint16_t minInterval, uint16_t maxInterval,
 } // updateConnParams
 
 
+/**
+ * @brief Request an update of the data packet length.
+ * * Can only be used after a connection has been established.
+ * @param [in] tx_octets The preferred number of payload octets to use (Range 0x001B-0x00FB).
+ * @param [in] tx_time The preferred number of microseconds to use when transmitting a single packet (Range 0x0148-0x4290).
+ */
+void NimBLEServer::setDataLen(uint16_t tx_octets, uint16_t tx_time) {
```
h2zero: This needs to be `NimBLEClient::setDataLen` here.

src/NimBLEServer.cpp
```diff
@@ -776,6 +776,21 @@ void NimBLEServer::updateConnParams(uint16_t conn_handle,
 }// updateConnParams
 
 
+/**
+ * @brief Request an update of the data packet length.
+ * * Can only be used after a connection has been established.
+ * @param [in] conn_handle The connection handle of the peer to send the request to.
+ * @param [in] tx_octets The preferred number of payload octets to use (Range 0x001B-0x00FB).
+ * @param [in] tx_time The preferred number of microseconds to use when transmitting a single packet (Range 0x0148-0x4290).
```
h2zero: Could you please add 
@details For more information about this function and data length extension see: https://punchthrough.com/maximizing-ble-throughput-part-3-data-length-extension-dle-2/ 

Also this in the client code.

h2zero (changes requested):
Looks good, just a couple notes to take care of. 

I'm thinking we should just calculate the `tx_time` using on the `(tx_octets` + 14) * 8;` to remove the need for the user to provide it. Thoughts?

Mr-Mime:
> I'm thinking we should just calculate the `tx_time` using on the `(tx_octets + 14) * 8;` to remove the need for the user to provide it. Thoughts?

Yeah definitely.
This also explains why 0x0849 was the maximum.

src/NimBLEClient.cpp
Mr-Mime: Whops, done.

src/NimBLEServer.cpp
Mr-Mime: Let me know if you are happy with the new @details.

If not, let me know what you had in mind :)

h2zero: I meant to put what I had in my comment above but yours is better 😄, thanks!

h2zero (approved):

h2zero:
Looks good to me! Thanks!

manhere:
` pServer->setDataLen(desc->conn_handle, 251);`
i use  in 'onConnect()' of class 'ServerCallbacks' ,but nothing changed.
any code example? 
thx!

manhere:
` pServer->setDataLen(desc->conn_handle, 251);`
i use  in 'onConnect()' of class 'ServerCallbacks' ,but nothing changed.
any code example? 
thx!

Mr-Mime:
Used it the same way.
Only difference is that I passed the 251 in hex format as `0x00fb`.

What exactly do you mean by "nothing changed"?
My application transfers data from the esp to an mobile phone, and I was able to see an increases data throughput, after setting the data length.

manhere:
> What exactly do you mean by "nothing changed"? My application transfers data from the esp to an mobile phone, and I was able to see an increases data throughput, after setting the data length.
Thank you for quick reply ,I used that for **notify**...but the payload data still was truncated to 20 bytes.
this is what i said “nothing change”, does this option have effect on **notify**?

Mr-Mime:
In the [docs](https://h2zero.github.io/esp-nimble-cpp/class_nim_b_l_e_characteristic.html#aa45461059e1992c816e32c371d17e813) it says that a notify only transmits 20 bytes.

For the complete characteristic value the other site needs to read from this characteristic, then the higher data length will be used.

> A notification is a transmission of up to the first 20 bytes of the characteristic value.

manhere:
opps~~ done by NimBLEDevice::setMTU(517)

joseitor:
> opps~~ done by NimBLEDevice::setMTU(517)

Have you been able to send more than 20 bytes by notify() with ESP32? Can you show me the code?

PR #286, opened 2021-09-15
Add NimBLEAttValue class.

h2zero:
This is a specialized container class to hold BLE attribute values.

- Removes the use of std::string previously used to store the values.
- Allows for setting/getting/notifying values from std::string, std::vector<uint8_t>, Arduino String, const char*, and uint8_t buffers.
- Has operators retrieve the value as std::string, Arduino String, std::vector<uint8_t>, uint8_t* or char pointers.
- Includes iterators and subscript/random access operator.
- Introduces a max length parameter to the creation of server characteristics/descriptors to limit the size of the memory footprint.
- Nearly Seamless integration with existing code.
- Adds a config option to enable/disable timestamp storage when the value is updated.
- Adds a config option to specify the initial size of the value container if not specified in the constructor.

### Writing POD types to a remote characteristic:
- `pRemoteChar->writeValue(10);`
- `pRemoteChar->writeValue("TEST");`
- `pRemoteChar->writeValue(23.45);`
- `pRemoteChar->writeValue('X');`

### Write container types to a remote characteristic:
- `pRemoteChar->writeValue(String("test string");`
- `pRemoteChar->writeValue(std::string("test std::string");`
- `pRemoteChar->writeValue(std::vector<uint8_t>({1,2,3,4,5,6}));`
- `pRemoteChar->writeValue(NimBLEAttValue({1,2,3,4,5,6}));`

### Reading POD types from a remote characteristic:
- `int i_val =  pRemoteChar->readValue<int>();`
- `double d_val =  pRemoteChar->readValue<double>();
- `const uint8_t* u8p = pRemoteChar->readValue();

### Reading container types from a remote characteristic:
- `std::string ss_str(pRemoteChar->readValue());`
- `String ar_str = pRemoteChar->readValue();`
- `std::vector<uint8_t> vec = pRemoteChar->readValue();`
- `NimBLEAttValue att_val = pRemoteChar->readValue();`

### Reading a single byte from the value:
`uint8_t val =  pRemoteChar->readValue()[2];`

### Printing the hex bytes of a value:
```
for (auto it :  pRemoteChar->readValue()) {
    printf("value: %02x\n", it);
 }
```

Note: The server attributes work the same way but with `setValue()`/`getVaue()`

### Directly using the NimBLEAttValue class:
```
// create an attribute value with a maximum capacity of 8 bytes, initialized to "TEST" (an extra byte will be allocated for the NULL) 
NimBLEAttValue my_att_val("TEST", 8); 

my_att_val += "1234"; // Value now contains "TEST1234"
printf("my_att_val = %s\n", my_att_val.c_str()); // use the c_str() method, print's "TEST1234"

// Concatenate a String and a NimBLEAttValue (same use for std::string)
String test_str = "my att value is: ";
test_str += my_att_val;
Serial.println(test_str);

```

### DIFF
--- a/src/NimBLEAttValue.h
+++ b/src/NimBLEAttValue.h
@@ -1,0 +1,447 @@
+/*
+ * NimBLEAttValue.h
+ *
+ *  Created: on March 18, 2021
+ *      Author H2zero
+ *
+ */
+
+#ifndef MAIN_NIMBLEATTVALUE_H_
+#define MAIN_NIMBLEATTVALUE_H_
+#include "nimconfig.h"
+#if defined(CONFIG_BT_ENABLED)
+
+#ifdef NIMBLE_CPP_ARDUINO_STRING_AVAILABLE
+#include <Arduino.h>
+#endif
+
+#include "NimBLELog.h"
+
+/****  FIX COMPILATION ****/
+#undef min
+#undef max
+/**************************/
+
+#include <string>
+#include <vector>
+
+#ifndef CONFIG_NIMBLE_CPP_ATT_VALUE_TIMESTAMP_ENABLED
+#    define CONFIG_NIMBLE_CPP_ATT_VALUE_TIMESTAMP_ENABLED 0
+#endif
+
+#if CONFIG_NIMBLE_CPP_ATT_VALUE_TIMESTAMP_ENABLED
+#    include <time.h>
+#endif
+
+#if !defined(CONFIG_NIMBLE_CPP_ATT_VALUE_INIT_LENGTH)
+#    define CONFIG_NIMBLE_CPP_ATT_VALUE_INIT_LENGTH 20
+#elif CONFIG_NIMBLE_CPP_ATT_VALUE_INIT_LENGTH > BLE_ATT_ATTR_MAX_LEN
+#    error CONFIG_NIMBLE_CPP_ATT_VALUE_INIT_LENGTH cannot be larger than 512 (BLE_ATT_ATTR_MAX_LEN)
+#elif CONFIG_NIMBLE_CPP_ATT_VALUE_INIT_LENGTH < 1
+#    error CONFIG_NIMBLE_CPP_ATT_VALUE_INIT_LENGTH cannot be less than 1; Range = 1 : 512
+#endif
+
+
+/* Used to determine if the type passed to a template has a c_str() and length() method. */
+template <typename T, typename = void, typename = void>
+struct Has_c_str_len : std::false_type {};
+
+template <typename T>
+struct Has_c_str_len<T, decltype(void(std::declval<T &>().c_str())),
+                     decltype(void(std::declval<T &>().length()))> : std::true_type {};
+
+
+/**
+ * @brief A specialized container class to hold BLE attribute values.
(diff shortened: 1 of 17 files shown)


src/NimBLEAttValue.h
```diff
@@ -0,0 +1,210 @@
...
+    time_t       m_timestamp = 0;
+
+    void         deepCopy(const NimBLEAttValue & source);
+
+public:
+    NimBLEAttValue(uint16_t init_len = NIMBLE_ATT_INIT_LENGTH, uint16_t max_len = BLE_ATT_ATTR_MAX_LEN);
+    NimBLEAttValue(const uint8_t *value, uint16_t len);
+    NimBLEAttValue(const char *value):NimBLEAttValue((uint8_t*)value, strlen(value)){}
+    NimBLEAttValue(const std::string str):NimBLEAttValue((uint8_t*)str.data(), str.length()){}
+    NimBLEAttValue(const NimBLEAttValue & source) { deepCopy(source); }
```
sivar2311, line 133: This might crash if `m_attr_value` is not initialized to `nullptr` (check in `deepCopy` will fail)

Suggestion: Initialize `m_attr_value`
```C++
NimBLEAttValue(const NimBLEAttValue & source) : m_attr_value(nullptr) { deepCopy(source); }
```

src/NimBLEAttValue.h
```diff
@@ -0,0 +1,210 @@
...
+    m_attr_value   = (uint8_t*)calloc(init_len + 1, 1);
+    assert(m_attr_value != nullptr && "No Mem");
+    m_attr_max_len = max_len;
+    m_attr_len     = 0;
+    m_timestamp    = 0;
+}
+
+inline NimBLEAttValue::NimBLEAttValue(NimBLEAttValue && source) {
+    m_attr_value   = source.m_attr_value;
+    m_attr_value   = source.m_attr_value;
```
sivar2311: Duplicate of line 92

src/NimBLEAttValue.h
```diff
@@ -0,0 +1,210 @@
...
+        free(m_attr_value);
+    }
+}
+
+inline NimBLEAttValue& NimBLEAttValue::operator =(NimBLEAttValue && source) {
+    if (this != &source){
+        free(m_attr_value);
+
+        m_attr_value   = source.m_attr_value;
+        m_attr_value   = source.m_attr_value;
```
sivar2311: Duplicate of line 109

src/NimBLEAttValue.h
h2zero: Good catch, thanks!

src/NimBLEAttValue.h
h2zero: Ditto here

src/NimBLEAttValue.h
h2zero: Nice find, noted!

h2zero:
Finally got around to cleaning this up, if anyone would like to test it out please provide feedback. I'd be especially interested in some criticism from any C++ template wizards out there.

h2zero:
Merging this as I have seen no issues so far in long term testing.

PR #304, opened 2021-10-16
Add optional clearing of scan results when stopping scan.

h2zero:
This resolves a possible collision when scanning is active and a call to connect to a device is made.

When the scanner is still active in a continuous scan the results may be pushed at the same time as clearing them in the call to stop scanning when attempting to connect to a device. This propses to add a parameter to the scan stop call to not clear the scan results in this scenario and may be useful for other use cases.


DTTerastar:
OK, i've made the `if (!pBLEScan->isScanning())` change and will report back on uptime

PR #330, opened 2022-01-10
Update version in library.properties

thorrak:
I'm having trouble pulling the latest version of the library automatically from platformio, and I think it relates to the version number not being updated in this file.


h2zero:
That is strange, I looked at platformio's registry and it's showing 1.3.1, seems to not be picking up the release branch.
https://github.com/h2zero/NimBLE-Arduino/tree/release/1.3.0

Issue #342, opened 2022-01-29
Request of setting notify_callback as class instance.


asukiaaa:
I tried to set `notify_callback` with using class member method but could not. (`scanCompleteCB` also)
Could you support `CharacteristicCallbacks` class to set notify_callback like `AdvertisedDeviceCallbacks` or `ClientCallbacks`?

Is there any way to do that current functions?

My use case:
```cpp
class Controller() {
 public:
  bool afterConnect(NimBLEClient* pClient) {
    for (auto pService : *pClient->getServices(true)) {
      auto sUuid = pService->getUUID();
      if (!sUuid.equals(uuidServiceHid)) {
        continue;  // skip
      }
      Serial.println(pService->toString().c_str());
      for (auto pChara : *pService->getCharacteristics(true)) {
        charaRead(pChara);
        charaSubscribeNotification(pChara);
      }
    }

    return true;
  }

  void charaSubscribeNotification(NimBLERemoteCharacteristic* pChara) {
    if (pChara->canNotify()) {
      charaPrintId(pChara);
      Serial.println(" canNotify ");
      if (pChara->subscribe(true, notifyCB, true)) { // <<<<<< ========== Here
        Serial.println("set notifyCb");
        // return true;
      } else {
        Serial.println("failed to subscribe");
      }
    }
  }

  void notifyCB(NimBLERemoteCharacteristic* pRemoteCharacteristic, uint8_t* pData,
                size_t length, bool isNotify) {
    // update controller state
  }
}

Controllelr myController;
```

Thank you for sharing an useful library.

sivar2311:
You can use `std::bind` to set a class member function as callback function:

```C++
  void charaSubscribeNotification(NimBLERemoteCharacteristic* pChara) {
    if (pChara->canNotify()) {
      charaPrintId(pChara);
      Serial.println(" canNotify ");
      if (pChara->subscribe(true, std::bind(&Controller::notifyCB, this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3, std::placeholders::_4), true)) { 
        Serial.println("set notifyCb");
      } else {
        Serial.println("failed to subscribe");
      }
    }
```

asukiaaa:
@sivar2311 
Thank you for the information.
I could call!

I tried calling scanEndCB with using the bind function but I cannot pass compilcation.
Could you give me some advice?
```cpp
class Controller {
  void startScan() {
    scanning = true;
[...]

PR #343, opened 2022-01-29
Define scanCompleteCB with using std::function

asukiaaa:
I want to use bind for scanCompleteCB like this.
```cpp
class Controller {
  void startScan() {
    scanning = true;
    auto pScan = NimBLEDevice::getScan();
    pScan->setAdvertisedDeviceCallbacks(advDeviceCBs);
    pScan->setInterval(45);
    pScan->setWindow(15);
    Serial.println("Start scan");
    // pScan->start(scanTime, scanEndedCB);
    pScan->start(scanTime,
                 std::bind(&Core::scanEndedCB, this, std::placeholders::_1)); // <<<<<<<<<< ======== Here
  }

  void scanEndedCB(NimBLEScanResults results) {
    Serial.println("Scan Ended");
    scanning = false;
  }
};
```

Related issue: https://github.com/h2zero/NimBLE-Arduino/issues/342

### DIFF
--- a/src/NimBLEScan.cpp
+++ b/src/NimBLEScan.cpp
@@ -282,7 +282,7 @@ bool NimBLEScan::isScanning() {
  * @param [in] is_continue Set to true to save previous scan results, false to clear them.
  * @return True if scan started or false if there was an error.
  */
-bool NimBLEScan::start(uint32_t duration, void (*scanCompleteCB)(NimBLEScanResults), bool is_continue) {
+bool NimBLEScan::start(uint32_t duration, scan_complete_callback scanCompleteCB, bool is_continue) {
     NIMBLE_LOGD(LOG_TAG, ">> start: duration=%" PRIu32, duration);
 
     // Save the callback to be invoked when the scan completes.
--- a/src/NimBLEScan.h
+++ b/src/NimBLEScan.h
@@ -27,6 +27,7 @@
 #endif
 
 #include <vector>
+#include <functional>
 
 class NimBLEDevice;
 class NimBLEScan;
@@ -55,6 +56,8 @@ private:
     std::vector<NimBLEAdvertisedDevice*> m_advertisedDevicesVector;
 };
 
+typedef std::function<void (NimBLEScanResults scanResults)> scan_complete_callback;
+
 /**
  * @brief Perform and manage %BLE scans.
  *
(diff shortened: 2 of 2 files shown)


h2zero:
Sorry for the delay. 

This is a good idea but we have a problem. 

This will break the API for users that do not provide the `is_continue` parameter when they provide a callback that does not use  std::bind (such as the examples). In this scenario the callback argument is converted to a `bool(true)` by the compiler and results in calling the wrong `start()` method, and the callback is not called.

In order to support this correctly another solution is needed. If we are going to break the API, I think the best way would be to have a `setScanEndCallback` method or a callback class added to `NimBLEScan` and remove the function parameter from the start calls. This would also require renaming the `start()` method to support the blocking and non-blocking calls with different naming.

I'm open to other suggestions, this is just my first thoughts.

Note: merging this will need to wait until the other breaking changes I have in mind are implemented with the upcoming 2.0 releases.

asukiaaa:
Thank you for the response.
I understand that this change breaks current code.

> have a setScanEndCallback method or a callback class added to NimBLEScan

I agree passing a class instance for a function because we can support additional callback easily.

> I have in mind are implemented with the upcoming 2.0 releases.

I see, thanks.

h2zero:
@asukiaaa Please take a look at #389, it is part of a few breaking changes I have planned for a version 2 release and should suit your needs.

asukiaaa:
Thank you for updating but it seems renaming m_pAdvertisedDeviceCallbacks to m_pScanCallbacks and reuse it as m_scanCompleteCB.
Is it possible to distinguish the timing of finding some device and finishing scanning?

h2zero:
Sorry, I need to update an examples to make it clearer. The advertised device callback class has been replaced with the scan callback class which contains the virtual function `onResult` and `onScanEnd` that you can define for your needs.

asukiaaa:
Thank you for the explanation.
I could use the updated NimBLE without using global functions and variables.
https://github.com/asukiaaa/arduino-XboxSeriesXControllerESP32/commit/57e50c181ff621cfd5c6521d0ea36bd6ab32c68d

h2zero:
You got it 😄

asukiaaa:
This request is satisfied by onScanEnd callback on https://github.com/h2zero/NimBLE-Arduino/pull/389.
Thank you.

PR #345, opened 2022-02-03
Fixes compile time error in PlatformIO conversion from 'uint8_t* to non-scalar

mc-hamster:
Fixes compile time error which happens in PlatformIO:

src/main.cpp: In member function 'virtual void DescriptorCallbacks::onWrite(NimBLEDescriptor*)':
src/main.cpp:155:47: error: conversion from 'uint8_t* {aka unsigned char*}' to non-scalar type 'std::__cxx11::string {aka std::__cxx11::basic_string<char>}' requested
     std::string dscVal = pDescriptor->getValue();

PR #356, opened 2022-02-16
changed getValue to getStringValue in example

sivar2311:
As mentioned by @mc-hamster in PR #345


h2zero:
This may not work in the future, I'm considering removing the `getStringValue` method as it is redundant in the master branch which uses the `NimBLEAttValue` class which provides an operator for this. I'm inclined to leave the current example code alone in this branch as it serves to verify the desired functionality.

Please test your code with the master branch without this change and let me know if you still have an issue.

sivar2311:
Ok, thanks. 
I did not recognize the different branches, sorry.
I was wondering why `NimBLEDescriptor` does not use the new `NimBLEAttrValue` class.

Issue #383, opened 2022-04-12
Remove first / from "/***...****/" comment lines in nimconfig.h


Bascy:
Current compilation doesnt like using `/****************/` as the last line of a block comment, complaining about

```
.pio/libdeps/debug/NimBLE-Arduino/src/nimconfig.h:38:2: warning: "/*" within comment [-Wcomment]
  /**************************************************/
  ^
```

PR #384, opened 2022-04-12
Removed /* from inside block comment(#383)

Bascy:


h2zero:
Oops, weird that the arduino IDE doesn't catch this. Thanks! Making a mental note to test with platformio before merging.

Issue #393, opened 2022-04-30
Version number in library.properties does not match release versions


Bascy:
Release 1.3.7 and 1.3.8 both contain version nr 1.3.6 in library.properties.
This can cause problems when using a dependency on those specific versions

i.e.
  dependency on https://github.com/h2zero/NimBLE-Arduino @ ^1.3.7 will raise an error as the downloaded lib with tag 1.3.7 contains 1.3.6 version label this will cause the following error:

> Error: Package version 1.3.6+sha.8b5cfb0 doesn't satisfy requirements ^1.3.7 based on PackageMetaData <type=library name=NimBLE-Arduino version=1.3.6+sha.8b5cfb0 spec={'owner': None, 'id': None, 'name': 'NimBLE-Arduino', 'requirements': '^1.3.7', 'uri': 'git+https://github.com/h2zero/NimBLE-Arduino'}

Bascy:
See PR #394

h2zero:
This is due to using a release branch for 1.3.X versions, if you use `h2zero/NimBLE-Arduino@^1.3.7` instead of pointing to the git repo it works fine. The master branch isn't part of any release yet so the version hasn't been updated. 

Thanks for the PR I will merge it to resolve this.

PR #394, opened 2022-04-30
Increased version nr in library.properties (#Issue #393)

Bascy:


library.properties
```diff
@@ -1,5 +1,5 @@
 name=NimBLE-Arduino
-version=1.3.6
+version=1.3.9
```
h2zero, line 2: Shouldn't this be 1.3.8?

Bascy: As 1.3.8 is already released, I would think this is set for the next release

h2zero: Next release will probably be 2.0.0, I think we should have the master set to the same version number as the release/1.3  branch for now to avoid confusion.

PR #395, opened 2022-04-30
[Breaking] Change all time input parameters to milliseconds.

h2zero:
Changes all functions that accept a time parameter to use milliseconds instead of seconds.

* Adds duration input to NimBLEDevice::startAdvertising and NimBLEServer::startAdvertising.


sivar2311:
Yippiiii  :partying_face:  (#388) 👍

PR #450, opened 2022-09-19
Update CHANGELOG.md to include config options

j4m3s:
I've added the changes that were made to the config options to the changelog.  Changes to these are backward-breaking for users that are specifying them via compile flags or env vars, but there's no way for people to know, so definitely worth listing here IMHO :)


h2zero:
Thanks! Sorry, I forgot about those.

PR #480, opened 2022-11-25
Clear duplicates cache when not saving scan results an no duplicates filtering

Bascy:
When running indefinite scan, results should not be stored (maxResults == 0), or at least be limited (maxResult < 0xFF) as this will keep increasing memory use.
But also the duplicates cache should be cleared after callback is invoked otherwise this cache will keep increasing.


h2zero:
I'm curious as to what prompted the need for this? The duplicate cache is maintained by the controller and clears itself when full and should allocate the storage at initialization.

Bascy:
I thought it was a solution to our memory problem, but lets putt his on hold for now ...

PR #504, opened 2023-02-11
Expose advertisement flags in NimBLEAdvertisedDevice

lultimouomo:
NimBLE allows settings advertisement data when advertising, but does not expose it when scanning.

### DIFF
--- a/src/NimBLEAdvertisedDevice.cpp
+++ b/src/NimBLEAdvertisedDevice.cpp
@@ -63,6 +63,25 @@ uint8_t NimBLEAdvertisedDevice::getAdvType() {
 
 
 /**
+ * @brief Get the advertisement flags.
+ * @return The advertisement flags, a bitmask of:
+ * BLE_HS_ADV_F_DISC_LTD (0x01) - limited discoverability
+ * BLE_HS_ADV_F_DISC_GEN (0x02) - general discoverability
+ * BLE_HS_ADV_F_BREDR_UNSUP - BR/EDR not supported
+ */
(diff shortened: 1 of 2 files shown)


src/NimBLEAdvertisedDevice.cpp
```diff
@@ -62,6 +62,25 @@ uint8_t NimBLEAdvertisedDevice::getAdvType() {
...
+ * @brief Get the advertisement flags.
+ * @return The advertisement flags, a bitmask of:
+ * BLE_HS_ADV_F_DISC_LTD (0x01) - limited discoverability
+ * BLE_HS_ADV_F_DISC_GEN (0x02) - general discoverability
+ * BLE_HS_ADV_F_BREDR_UNSUP - BR/EDR not supported
+ */
+uint8_t NimBLEAdvertisedDevice::getAdvFlags() {
+    size_t data_loc = 0;
+
+    if(findAdvField(BLE_HS_ADV_TYPE_APPEARANCE, 0, &data_loc) > 0) {
```
h2zero: I think this should be `BLE_HS_ADV_TYPE_FLAGS` instead of `BLE_HS_ADV_TYPE_APPEARANCE`.

h2zero (changes requested):

src/NimBLEAdvertisedDevice.cpp
lultimouomo: You're right, fixed

h2zero (approved):

PR #507, opened 2023-02-20
Add directed peer address parameter to advertising start.

### LINKED ISSUE #501: peer_addr = null when doing directed advertising
When I do "BLE_GAP_CONN_MODE_UND" everything works but when I change to "BLE_GAP_CONN_MODE_DIR" I always get the error "Unable to advertise - Duration too long". I have found out that this is a result of the rc value of 3 (BLE_HS_EINVAL). This error code is used in a lot of places and I started replacing the values with my own unique error codes and found out that my error was peer_addr being null in the file ble_gap.c in the function ble_gap_adv_validate. But i cant fix this since the file NimBLEAdvertising.cpp puts null in the call ble_gap_adv_start. 

TLDR: NimBLEAdvertising.cpp sets peer_addr to null and ble_gap.c returns an error if peer_addr is null

h2zero:
Adds a parameter to NimBLEAdvertising::start to take the peer address for use with directed advertising.

Fixes #501

### DIFF
--- a/src/NimBLEAdvertising.cpp
+++ b/src/NimBLEAdvertising.cpp
@@ -387,9 +387,10 @@ void NimBLEAdvertising::setScanResponseData(NimBLEAdvertisementData& advertiseme
  * @brief Start advertising.
  * @param [in] duration The duration, in seconds, to advertise, 0 == advertise forever.
  * @param [in] advCompleteCB A pointer to a callback to be invoked when advertising ends.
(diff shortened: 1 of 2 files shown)


h2zero:
@ViktorRindom Have you been able to test this?

ViktorRindom:
Sorry but no. my implementation is using undirected connections now because i have to accept an entire whitelist of devices. i dont have time to make a example to test it but thank you for the quick fix.

PR #520, opened 2023-03-26
Fix CONFIG_BT_NIMBLE_NVS_PERSIST value not being used.

### LINKED ISSUE #518: Value of CONFIG_BT_NIMBLE_NVS_PERSIST is ignored by compiler
In esp_nimble_cfg.h the existence of a #define for `CONFIG_BT_NIMBLE_NVS_PERSIST` is checked, but the value is ignored, so (at least for my compiler) this always sets `MYNEWT_VAL_BLE_STORE_CONFIG_PERSIST` to 1.
I am not totally sure, if this is undefined behavior and may work for some compilers or if it is completely wrong.

For me this change works:
```
#ifndef MYNEWT_VAL_BLE_STORE_CONFIG_PERSIST
#if     CONFIG_BT_NIMBLE_NVS_PERSIST==1
#define MYNEWT_VAL_BLE_STORE_CONFIG_PERSIST (1)
#else
#define MYNEWT_VAL_BLE_STORE_CONFIG_PERSIST (0)
#endif
#endif
```

h2zero:
Fixes #518


Staars:
Thanks.

PR #529, opened 2023-04-17
Set manufacturer data directly instead of converting from string

eighty2fifty1:
For my purposes, I found it useful to be able to directly input integers into manufacturer data, so I created this PR to allow me to do so. It works perfectly on my application (ESP32 DEVKIT V1).


h2zero:
Thank you for this, I will merge it soon.

h2zero:
@eighty2fifty1 I made some minor changes if you'd like to comment/test before merge?

PR #584, opened 2023-08-19
Fix NimBLEDevice::init() deadlock problem

CW-B-W:
`NimBLEDevice::host_task()` has priority `(configMAX_PRIORITIES - 4)`, which is defined at 
https://github.com/h2zero/NimBLE-Arduino/blob/bc333ccb6e8d9ff2059af9cbd5006a290a4de3a5/src/nimble/porting/npl/freertos/src/nimble_port_freertos.c#L60-L61

When the priority of `NimBLEDevice::init()` is greater than `(configMAX_PRIORITIES - 4)`, `NimBLEDevice::init()` will block `NimBLEDevice::host_task()` but still wait for `NimBLEDevice::host_task()` infinitely (waiting for `m_synced` to be set), resulting in the deadlock between `NimBLEDevice::init()` and `NimBLEDevice::host_task()`.

This is caused by this while loop
https://github.com/h2zero/NimBLE-Arduino/blob/bc333ccb6e8d9ff2059af9cbd5006a290a4de3a5/src/NimBLEDevice.cpp#L909-L911
When the priority of `NimBLEDevice::init()` is greater than `NimBLEDevice::host_task()`, it cannot yield and let `NimBLEDevice::host_task()` be executed.

Using a binary semaphore can resolve this problem.

### DIFF
--- a/src/NimBLEDevice.cpp
+++ b/src/NimBLEDevice.cpp
@@ -68,6 +68,7 @@ NimBLEServer*   NimBLEDevice::m_pServer = nullptr;
 #endif
 uint32_t        NimBLEDevice::m_passkey = 123456;
 bool            NimBLEDevice::m_synced = false;
+SemaphoreHandle_t m_synced_sem = xSemaphoreCreateBinary();
 #if defined(CONFIG_BT_NIMBLE_ROLE_BROADCASTER)
 #  if CONFIG_BT_NIMBLE_EXT_ADV
 NimBLEExtAdvertising* NimBLEDevice::m_bleAdvertising = nullptr;
@@ -807,6 +808,7 @@ void NimBLEDevice::onSync(void)
     taskYIELD();
 
     m_synced = true;
+    xSemaphoreGive(m_synced_sem);
 
     if(initialized) {
 #if defined(CONFIG_BT_NIMBLE_ROLE_OBSERVER)
@@ -906,6 +908,7 @@ void NimBLEDevice::init(const std::string &deviceName) {
     }
 
     // Wait for host and controller to sync before returning and accepting new tasks
+    xSemaphoreTake(m_synced_sem, portMAX_DELAY);
     while(!m_synced){
         taskYIELD();
     }


h2zero:
I've never heard of or experienced this issue before, do you have any examples where this can be reproduced?

h2zero:
Due to the nature of the BLE stack I would highly recommend using a task priority for the application that is less than the host task. Is there a specific purpose to have your application run at a higher priority than the host task?

CW-B-W:
> Due to the nature of the BLE stack I would highly recommend using a task priority for the application that is less than the host task. Is there a specific purpose to have your application run at a higher priority than the host task?

Actually, I don't really need the priority of application be higher, I just accidentally found this problem when I set different task priority for my application.
I was porting this project to an MCU (other than Arduino, ESP32), and it uses polling to read user command from UART, one of the UART command is to turn BLE on.
To make the UART command handler be as responsive as possible, I made its priority the highest `configMAX_PRIORITIES - 1`, then I found when I launch `NimBLEDevice::init()` inside this command handler, which has the highest priority, it gets stuck.

h2zero:
@CW-B-W That is a great explanation, thank you for this. I have tested your example and propose a different fix. 

Instead of the semaphore, could you just replace the `taskYIELD()` in the `while` loop with `ble_npl_time_delay(1)` instead, that should resolve the issue as well without the extra memory requirements.

i-am-shodan:
FYI I tried to use `ble_npl_time_delay(1)` I even went so far as making m_synced volatile. Doesn't work for me

i-am-shodan:
Doesn't look like the patch works for me either. xSemaphoreTake never returns.

h2zero:
@i-am-shodan you shouldn't need to do anything other that add a call to `delay()` in your own tasks/loop, this allows for a context switch.

xgnata:
I'm also experiencing crashes due to starvation of the host task.
I have to paly with the priority of several tasks in my code to make it "work", but it is fragile. 
I'm working on a reduced test case.

Is there a reason not to merge thechange proposed in https://github.com/h2zero/NimBLE-Arduino/pull/584#issuecomment-1826194526 ?? This change looks good to me.

h2zero:
Unless you have a very good reason to have any tasks with priority higher than the BLE host task then you should make sure that everything else is lower priority.

> Is there a reason not to merge thechange proposed in https://github.com/h2zero/NimBLE-Arduino/pull/584#issuecomment-1826194526 ?? This change looks good to me.

No reason, just haven't got to it yet 😄

xgnata:
Why is it so? Because the BLE stack is full of timers/watchdogs which will
trigger an error if they are not rested on time?

Xavier

h2zero:
Yes, there is a lot of things going on in the BLE stack and it needs time to process.

h2zero:
Closing this as the issue was resolved in #738

PR #609, opened 2023-11-25
Set a timer for scan responses.

h2zero:
This sets a timer for scan responses such that if they are not received within 1 second, the scan result callback will be called and the device deleted from the vector when max_results = 0.


h2zero:
Not confident with this fix, too much involved to implement without errors.

PR #654, opened 2024-04-03
Added nimble_port_freertos_get_hs_hwm() visibility for ESP32 platform

devmirek:
Added ability to track nimble_host task free stack for ESP32 and fixed wrong values in case of uninitialized BLE stack


h2zero:
Thanks!

PR #737, opened 2024-10-31
doc: mention 0 is forever for NimBLESCan::start

twilly:
Document that setting `duration` to zero means to scan forever.


twilly:
Super minor PR to add documentation that 0 is forever with NimBLEScan::start. I was confused why NimBLEAdvertising::start would be different, but it's not different at all. :)

h2zero:
Thanks!

PR #743, opened 2024-11-03
v1.4.2: Allow dynamically change advertising data

arkhipenko:
This small change allows dynamically changing advertising data without a stop/start cycle. 
I found it useful for changing the manufacturing data (for instance) frequently without affecting device's ability to be connected to or discovered. 
`refreshAdvertisingData()` should be called after methods that alter the data, e.g. `setManufacturerData(..)`

### DIFF
--- a/src/NimBLEAdvertising.cpp
+++ b/src/NimBLEAdvertising.cpp
@@ -386,6 +386,22 @@ void NimBLEAdvertising::clearData() {
 } // clearData
 
 /**
+ * @brief Refresh advertsing data dynamically without stop/start cycle.
+ * For instance allows refreshing manufacturer data dynamically.
+ *
+ * @return True if the data was set successfully.
+ * @details If scan response is enabled it will be refreshed as well.
+ */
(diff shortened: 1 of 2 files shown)


h2zero:
@arkhipenko Thank you for this, its a great idea. One thing that is missing is setting the scan response data to compliment this.

arkhipenko:
Not quite sure what you mean by this, but maybe.... 
I came across this need because advertising data is a great way to communicate out without the need to be connected. Like presenting a sensor data or something along those lines. How does scan response come into play here?

mhaberler:
I wasnt aware the response to an active scan might be different than the content of a periodic advertisement - good point

very much looking forward this capability!

arkhipenko:
Ok, I read up on it and figured what this is about. I just didn't need the scan response, so didn't think about it.  Yes, that capability would be a great compliment to adv. data. 
Thanks. 
I learn every day!

h2zero:
@arkhipenko I will be adding this after #756 is merged. I will make a couple edits to account for this event, just wanted to let you know.

PR #756, opened 2024-11-19
[BREAKING] - Refactor NimBLEAdvertising

### LINKED ISSUE #646: Set Manufacturer Data - Bug
Current implementation to set manufacturer data `void NimBLEAdvertising::setManufacturerData(const std::string &data)`

Issue: Trying to set the first two bytes of the manufacturer data to Manufacturer ID from Bluetooth SIG list (Nordic Semi ASA - 0x0059) results in no manufacturer data shown in advertisement

Current implementation:
```
void NimBLEAdvertising::setManufacturerData(const std::string &data)
{
    m_mfgData.assign(data.begin(), data.end());
    m_advData.mfg_data = &m_mfgData[0];
    m_advData.mfg_data_len = m_mfgData.size();
    m_advDataSet = false;
} // setManufacturerData
```

My example where it breaks:
```
uint16_t manufacturerID = 0x0059; // Nordic Semi ASA
uint8_t dataToAdvertise[4];
dataToAdvertise[0] = (manufacturerID >> 8) & 0xFF;
dataToAdvertise[1] = (manufacturerID & 0xFF);
dataToAdvertise[2] = 0xAB; // Placeholder random data
dataToAdvertise[3] = 0xCD; // Placeholder random data
pAdvertising->setManufacturerData((char *)dataToAdvertise); // Cast to char * to work with std::string 
```
m_mfgData.size() returns zero since the first byte in dataToAdvertise array is a null-character (0)

Proposed Alternative:
```
void NimBLEAdvertising::setManufacturerData(uint8_t *data, uint8_t length)
{
    m_mfgData.assign(data, data + length);
    m_advData.mfg_data = &m_mfgData[0];
    m_advData.mfg_data_len = length;
    m_advDataSet = false;
} // setManufacturerData
```
We can overload the setManufacturerData function with the uint8_t* type and a length parameter to update the advertising vector

```
uint16_t manufacturerID = 0x0059; // Nordic Semi ASA
uint8_t dataToAdvertise[4];
dataToAdvertise[0] = (manufacturerID >> 8) & 0xFF;
dataToAdvertise[1] = (manufacturerID & 0xFF);
dataToAdvertise[2] = 0xAB;
dataToAdvertise[3] = 0xCD;
pAdvertising->setManufacturerData(dataToAdvertise, sizeof(dataToAdvertise));
```

h2zero:
* General code cleanup.
* `NimBLEAdvertisementData` moved to it's own .h and .cpp files.
* Added new method, `NimBLEAdvertising::setPreferredParams` that takes the min and max preferred connection parameters as an alternative for `setMinPreferred` and `setMaxPreferred`.
* Added new method, `NimBLEAdvertising::setAdvertisingInterval` Sets the advertisement interval for min and max to the same value instead of calling `setMinInterval` and `setMaxInterval` separately if there is not value difference.
* `NimBLEAdvertisementData` payload is now stored in `std::vector<uint8_t>` instead of `std::string`.
* `NimBLEAdvertisementData::getPayload` now returns `std::vector<uint8_t>` instead of `std::string`.
* `NimBLEAdvertisementData::addData` now takes either a `std::vector<uint8_t>` or `uint8_t* + length` instead of `std::string` or `char + length`.
* `NimBLEAdvertisementData::setName` now takes an optional `bool` parameter to indicate if the name is complete or incomplete, default = complete.
* `NimBLEAdvertising::start` No longer takes a callback pointer parameter, instead the new method `NimBLEAdvertising::setAdvertisingCompleteCallback` should be used.
* `NimBLEAdvertising::setAdvertisementType` has been renamed to `NimBLEAdvertising::setConnectableMode` to better reflect it's function.
* Added new method, `NimBLEAdvertising::setDiscoverableMode` to allow applications to control the discoverability of the advertiser.
* Advertising the name and TX power of the device will no longer happen by default and should be set manually by the application.
* Added overload for `NimBLEAdvertising::setManufacturerData` that takes a `const uint8_t*` and , size_t` paramter.
* Added overload for `NimBLEAdvertising::setServiceData` that takes `const NimBLEUUID& uuid`, ` const uint8_t* data`, ` size_t length` as parameters.
* Added overload for `NimBLEAdvertising::setServiceData` that takes `const NimBLEUUID& uuid`, `const std::vector<uint8_t>&` as parameters.
* All `NimBLEAdvertisementData` functions that change data values now return `bool`, true = success.
* All `NimBLEAdvertising` functions that change data values now return `bool`, true = success.
* `NimBLEAdvertising::setMinPreferred` and `NimBLEAdvertising::setMaxPreferred` have been removed, use `NimBLEAdvertising::setPreferredParams` instead.
* All advertising data is now stored in instances of `NimBLEAdvertisingData` and vectors removed from `NimBLEAdvertising`.
* `NimBLEAdvertising::setAdvertisementData` and `NimBLEAdvertising::setScanResponseData` now return `bool`, true = success.
* Added new method, `NimBLEAdvertisementData::removeData`, which takes a parameter `uint8_t type`, the data type to remove.
* Added new method, `NimBLEAdvertisementData::toString`, which will print the data in hex.
* Added new method, `NimBLEAdvertising::getAdvertisementData`, which returns a reference to the currently set advertisement data.
* Added new method, `NimBLEAdvertising::getScanData`, which returns a reference to the currently set scan response data.
* Added overloads for `NimBLEAdvertising::removeServiceUUID` and `NimBLEAdvertisementData::removeServiceUUID` to accept a `const char*`
* Added new method, `NimBLEAdvertising::clearData`, which will clear the advertisement and scan response data.

Fixes #646

### DIFF
--- a/examples/BLE_EddystoneTLM_Beacon/BLE_EddystoneTLM_Beacon.ino
+++ b/examples/BLE_EddystoneTLM_Beacon/BLE_EddystoneTLM_Beacon.ino
@@ -12,7 +12,7 @@
    4. wait
    5. Stop advertising.
    6. deep sleep
-   
+
 */
 
 #include "NimBLEDevice.h"
@@ -26,7 +26,7 @@
 #define GPIO_DEEP_SLEEP_DURATION 10     // sleep x seconds and then wake up
 
 // UUID 1 128-Bit (may use linux tool uuidgen or random numbers via https://www.uuidgenerator.net/)
-#define BEACON_UUID "8ec76ea3-6668-48da-9866-75be8bc86f4d" 
+#define BEACON_UUID "8ec76ea3-6668-48da-9866-75be8bc86f4d"
 
 RTC_DATA_ATTR static time_t last;    // remember last boot in RTC Memory
 RTC_DATA_ATTR static uint32_t bootcount; // remember number of boots in RTC Memory
@@ -96,6 +96,7 @@ void setup()
   BLEDevice::setPower(ESP_PWR_LVL_N12);
 
   pAdvertising = BLEDevice::getAdvertising();
+  pAdvertising->enableScanResponse(true);
 
   setBeacon();
   // Start advertising
--- a/examples/BLE_EddystoneURL_Beacon/BLE_EddystoneURL_Beacon.ino
+++ b/examples/BLE_EddystoneURL_Beacon/BLE_EddystoneURL_Beacon.ino
@@ -26,7 +26,7 @@
 #define GPIO_DEEP_SLEEP_DURATION 10     // sleep x seconds and then wake up
 
 // UUID 1 128-Bit (may use linux tool uuidgen or random numbers via https://www.uuidgenerator.net/)
-#define BEACON_UUID "8ec76ea3-6668-48da-9866-75be8bc86f4d" 
+#define BEACON_UUID "8ec76ea3-6668-48da-9866-75be8bc86f4d"
 
 RTC_DATA_ATTR static time_t last;    // remember last boot in RTC Memory
 RTC_DATA_ATTR static uint32_t bootcount; // remember number of boots in RTC Memory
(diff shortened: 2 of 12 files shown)


h2zero:
This one is huge, if anyone would like to test I'd really like the feedback!

PR #767, opened 2024-11-29
[BREAKING] - Refactor NimBLEHIDDevice

h2zero:
* General code cleanup.
* All functions that return a characteristic or service have been renamed to have a `get` prefix to better represent their function.
* All functions that set a value have been renamed with the prefix `set` to better represent their functionality.
* Added an `bool notify` parameter to `setBatteryLevel`, which if `true`, will send a notification of the value to the peer.


lemmingDev:
@Sab1e - If you've got the time

lemmingDev:
> @lemmingDev, would you mind testing/reviewing this?

Hi @h2zero 

The latest NimBLE-Arduino 2.1.3 is working with ESP32-BLE-Gamepad after a couple of users added some patches to the library

One issue we seem to have now is that on Windows (tested on Windows 10 and 11), it works fine when you pair it the first time, however when you power the board down and it automatically reconnects, gamepad input no longer works.

The main loop seems to be working as any serial output or LEDs modified in the sketch still update.

Android and perhaps Linux seem to work fine still

If you've got the time; any advice on what might need to be changed?

There is a thread with some testing feedback here: https://github.com/lemmingDev/ESP32-BLE-Gamepad/issues/251

Thanks

h2zero:
Thanks @lemmingDev, could you please open an issue about this?

PR #801, opened 2024-12-15
Fix compiling errors on nRF devices

mr258876:
- Fixed "'class NimBLEServer' has no member named 'm_pClient'" when CONFIG_BT_NIMBLE_ROLE_CENTRAL not defined
- Fixed missing nimble_port_stop(void)


h2zero:
Thanks for this, I guess the CI doesn't catch everything 😄. Do you actually call `deinit` on your nRF device? I'm curious as to the reason.

PR #805, opened 2024-12-15
Fix NimBLEHIDDevice output report returning incorrect characteristic.

### LINKED ISSUE #804: Output Report Function Not Working Properly in Release 2.1.0
Hello @h2zero ,
I am a contributor to the ESP32-BLE-Gamepad repository. We use your repository as a dependency, and I must say that your project is excellent! However, when adapting to Release 2.1.0, all other features work fine, but the Output Report doesn't seem to work properly. Specifically, when the host computer sends data, the following issue occurs:
```
Output Report send failed
Error writing report. Code: 1168, Message: Element not found.
```
You can find the host application here: [HIDOutputDemo_CSharp](https://github.com/Sab1e-GitHub/HIDOutputDemo_CSharp)
On the other hand, with the NimBLE Arduino version 1.4.3, I can use all the features without issues. I believe I have carefully checked the code, and it seems the issue is not with our repository. 

Additionally, I have tried printing to check if the OutputReportChr pointer is correctly obtained, and it looks fine:
```
    if (enableOutputReport) {
    BleGamepadInstance->outputGamepad = BleGamepadInstance->hid->getOutputReport(BleGamepadInstance->configuration.getHidReportId());
    BleGamepadInstance->outputReceiver = new BleOutputReceiver(outputReportLength);
    if (BleGamepadInstance->outputGamepad == nullptr) {
        Serial.println("Failed to create Output Report characteristic!");
        return;
    }
    BleGamepadInstance->outputGamepad->setCallbacks(BleGamepadInstance->outputReceiver);

    Serial.println("Enabled Output Report!");
    }
```
图片

Here is the code I've modified in our repository: [ESP32-BLE-Gamepad/tree/master](https://github.com/Sab1e-GitHub/ESP32-BLE-Gamepad/tree/master)
And here is the old version of the repository code: [ESP32-BLE-Gamepad/tree/NimBLE-v1.4.3](https://github.com/Sab1e-GitHub/ESP32-BLE-Gamepad/tree/NimBLE-v1.4.3)

I am not very familiar with the Bluetooth low-level architecture, so I would greatly appreciate your help. Thank you very much!

h2zero:
The input and output report characteristics share the same UUID and the `getOutputReport` was returning the input report characteristic. This change will return the correct characteristic by finding it with the index parameter.

Fixes #804

### DIFF
--- a/src/NimBLEHIDDevice.cpp
+++ b/src/NimBLEHIDDevice.cpp
@@ -176,9 +176,11 @@ NimBLECharacteristic* NimBLEHIDDevice::getInputReport(uint8_t reportId) {
  * @param [in] reportId Output report ID, the same as in report map for output object related to the characteristic.
  * @return A pointer to the output report characteristic.
  * @details This will create the characteristic if not already created.
+ * @note The output report characteristic is optional and should only be created after the input report characteristic.
  */
 NimBLECharacteristic* NimBLEHIDDevice::getOutputReport(uint8_t reportId) {
-    NimBLECharacteristic* outputReportChr = m_hidSvc->getCharacteristic(inputReportChrUuid);
+    // Same uuid as input so this needs to be the second instance
+    NimBLECharacteristic* outputReportChr = m_hidSvc->getCharacteristic(inputReportChrUuid, 1);
     if (outputReportChr == nullptr) {
         outputReportChr =
             m_hidSvc->createCharacteristic(inputReportChrUuid,


h2zero:
Hi @afpineda, thanks for the feedback.
>You can have more than one output report.

I did not know this. 

The PR was only to address the issue brought up in #804, where it was confirmed to resolve the issue.

I'm starting to think that the best fix is to move this class into it's own repo/library as I know nothing of HID specs and have no use or interest for this myself. The only reason this is here is because it was in the original library so I converted it to use NimBLE. 
If you would like to open a repo and host this I would be happy to provide a link to it from here instead of hosting the code.

h2zero:
Not discouraged, just a bit weary. I get in trouble any time I touch this class lol. Like I said this is not my area of knowledge, just thought it would be better if someone that knows more about it maintained it.

PR #812, opened 2024-12-17
Fix getOutputReport, getInputReport and getFeatureReport: not being able to create more than one report + temporal coupling

afpineda:
This fix is related to Issue #805 in the `NimBLEHIDDevice` class.

*What fixes*:

- The user can call `getXXXXXReport()` in any order.
- The user gets a different characteristic for each different reportID.
- The user gets the same characteristic for the same reportID.

*Fix description*:

A new private method is introduced: `locateReportCharacteristicById()`: given a `reportID`, it returns the characteristic if already created, nullptr otherwise.
`getXXXXXReport()` tries to locate the characteristic. If found, it is returned "as is". If not found, it is created and returned.

*Drawbacks*:
There is a computational overhead. The user should store/cache the value returned by `getXXXXXReport()` instead of repeating the calls. Perhaps we should warn about this in the documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 docs/doxydocs
+.development
--- a/src/NimBLEHIDDevice.cpp
+++ b/src/NimBLEHIDDevice.cpp
@@ -157,18 +157,15 @@ void NimBLEHIDDevice::setBatteryLevel(uint8_t level, bool notify) {
  * @return NimBLECharacteristic* The characteristic.
  * @return nullptr If the characteristic does not exist.
  */
-NimBLECharacteristic* NimBLEHIDDevice::locateReportCharacteristicById(uint8_t reportId, uint8_t &reportType)
-{
-    NimBLECharacteristic* candidate = m_hidSvc->getCharacteristic(inputReportChrUuid,0);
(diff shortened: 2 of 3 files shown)


h2zero:
https://github.com/h2zero/NimBLE-Arduino/pull/805#issuecomment-2548600648

.gitignore
```diff
@@ -1 +1,2 @@
 docs/doxydocs
+.development
```
h2zero, line 2: Can we remove this?

h2zero (changes requested):
Looks good, just a minor comment. Also can you please check the formatting, if in vscode you can just right click and format document.

.gitignore
afpineda: > Can we remove this?

Sorry, no. Arduino does not allow you to edit a library unless the ".development" file is in the library folder, but at the same time it does not allow you to include the file in the repo.
See the [Arduino library specification](https://docs.arduino.cc/arduino-cli/library-specification/), section "Development flag file".
Note that I'm using a clone of your repo as a library for my own sketches.

It does no harm to anybody.

h2zero:
Thank you @afpineda !

PR #833, opened 2025-01-03
Allow the same report ID in multiple input/output/feature reports

afpineda:
Related to issue [279 in esp-nimble-cpp](https://github.com/h2zero/esp-nimble-cpp/issues/279).

What fixes:
- The user gets a different characteristic for different report IDs **or** report types.
- The user gets the same characteristic for the same report ID **and** report type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 docs/doxydocs
+.development
--- a/src/NimBLEHIDDevice.cpp
+++ b/src/NimBLEHIDDevice.cpp
@@ -150,21 +150,21 @@ void NimBLEHIDDevice::setBatteryLevel(uint8_t level, bool notify) {
 } // setBatteryLevel
 
 /**
- * @brief Locate the characteristic for a report ID.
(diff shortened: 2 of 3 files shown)


h2zero:
> Hope it work works just the same.

Sadly no, it actually creates 9 commits in this case instead of 1.

> Git rebase scares me a lot !!!.

It's only scary the first time 😄, it's actually quite easy and safe if you try it locally and don't push anything until it looks right, can always reset to remote if it goes wrong.

I'll just squash this for you in this case though as there are no conflicts that way.

PS: Another way to rebase safely without the rebase command is to create a patch file from the diff, reset --hard origin/master, then apply your patch file. Github will even create a patch for you like this: https://github.com/h2zero/NimBLE-Arduino/pull/833.patch

h2zero:
Closing this as it has been rebased and merged in [9864abe](https://github.com/h2zero/NimBLE-Arduino/commit/9864abe311cc61c12c67ca1e92c8e80ccab68f2e) Thanks!

PR #834, opened 2025-01-03
avoid unused-variable warning

mazgch:
Avoid warning when compiled with All warnings enabled.  

```
.../src/NimBLEScan.cpp: In member function 'void NimBLEScanResults::dump() const': 
.../src/NimBLEScan.cpp:481:22: warning: unused variable 'dev' [-Wunused-variable]
  481 |     for (const auto& dev : m_deviceVec) {
      |                      ^~~
```
      
  alternatively the #if could even wrap the whole for loop ...

### DIFF
--- a/src/NimBLEScan.cpp
+++ b/src/NimBLEScan.cpp
@@ -478,12 +478,11 @@ void NimBLEScan::clearResults() {
  * @brief Dump the scan results to the log.
  */
 void NimBLEScanResults::dump() const {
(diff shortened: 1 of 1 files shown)


h2zero:
Thanks! I thought I had fixed this already but I guess it didn't get merged. Could you change this to just wrap the loop with `#if CONFIG_NIMBLE_CPP_LOG_LEVEL >=3` please?

h2zero:
Thanks!

PR #836, opened 2025-01-04
Fix incorrect TX power setting for advertising.

### LINKED ISSUE #835: Wrong dbM value in advertisement data when using C3/S3 chip.
C3 and S3 have a workaround where different power levels could be set for e.g. advertising or scanning. The workaround does not take into account advertisement data, which keeps sending "default" channel with incorrect dbM value to the advertisement packet.

The fix should be backwards compatible since it only adds an optional parameter when Tx level is added into the advertisement packet.

Tested and verified that I now see -3 dbM when previously setting the Tx Level like this:

NimBLEDevice::setPowerLevel(ESP_PWR_LVL_N3, ESP_BLE_PWR_TYPE_ADV);

Without the fix some default value is transmitted.

h2zero:
* Adds a new enum class to specify the tx power type to set (unused by non-esp devices).
* When all type is specified it is now ensured that all the tx power types are set to the specified level.
* Correctly sets the power level advertised for extended advertising with legacy pdu selected.

Fixes #835


h2zero:
@sanastasiou

Issue #837, opened 2025-01-04
Setting advertisement of custom UUID's are funky


doudar:
I have a cycling app that uses some custom generated UUID's that are not advertised after adding them to advertising.

This is using NimBLE-Arduino 2.1.2 and Platform 6.9.0

I've added and removed them (One at a time) for testing and here are my findings:

pAdvertising->addServiceUUID(NimBLEUUID((uint16_t)0x180D)); **//Works**

pAdvertising->addServiceUUID(NimBLEUUID("b4cc1223-bc02-4cae-adb9-1217ad2860d1")); **//No work :(**

pAdvertising->addServiceUUID(NimBLEUUID("77776277-7877-7774-4466-896665500000")); **//No work :(**

//Interestingly, If I do
NimBLEService *pService = spinBLEServer.pServer->createService(NimBLEUUID("b4cc1223-bc02-4cae-adb9-1217ad2860d1"));
spinBLEServer.pServer->getAdvertising()->addServiceUUID(pService->getUUID()); **//No Work**

NimBLEService *pService = spinBLEServer.pServer->createService(NimBLEUUID("77776277-7877-7774-4466-896665500000"));
spinBLEServer.pServer->getAdvertising()->addServiceUUID(pService->getUUID()); **//Works**

I didn't have this issue with the 1.x version of NimBLE (but it could be a platform bug too...)

h2zero:
Happy New Year @doudar I'll have a look, strange indeed.

h2zero:
@doudar I think I see your issue, make sure you enable scan response before adding anything to the advertising and it should work.

doudar:
Thanks! Heading to dinner, I'll let you know if if that works when i get back! 

Happy to see you working on 2.0!

doudar:
@h2zero , Enabling scan response BEFORE adding advertising seemed to fix the Advertised name not being set as well as being able to add the UUID to advertising without doing the workaround :

``pServer->getAdvertising()->addServiceUUID(pService->getUUID());```

I still cannot get this UUID to add to advertising, no matter how I set it. Give it a try on your end:
pServer->getAdvertising()->addServiceUUID(NimBLEUUID("b4cc1223-bc02-4cae-adb9-1217ad2860d1"));

h2zero:
@doudar that's the code I used to test the related PR #838. Could you try your code using that branch and let me know?

doudar:
[...]

PR #841, opened 2025-01-05
NimBLEAdvertisementData emit error on failure, remove magic numbers.

thekurtovic:
https://github.com/h2zero/NimBLE-Arduino/issues/837


thekurtovic:
Wanted to clean up cases like this `addData(data, 3)` to `addData(data, sizeof(data))` but there are quite a few so it's probably best to leave that for a separate style/convention PR.

h2zero:
LGTM, Thanks!

> Wanted to clean up cases like this addData(data, 3) to addData(data, sizeof(data)) but there are quite a few so it's probably best to leave that for a separate style/convention PR.

Yeah that definitely should be a separate PR.

Issue #884, opened 2025-02-04
Weird crash with discoverAttributes() in 2.2.1


gkoh:
I was updating to the latest and greatest 2.2.1 to verify both #861 and #867 when I got a consistent crash.
Using the same code as per #861:
https://github.com/h2zero/NimBLE-Arduino/issues/861#issue-2788589893

and trying to connect to whatever happens to allow it in the area, things seem to go horribly wrong:
```
D NimBLERemoteService: >> retrieveCharacteristics()
I (2733) NimBLE: GATT procedure initiated: discover all characteristics; 
I (2740) NimBLE: start_handle=1 end_handle=5

D NimBLERemoteService: Characteristic Discovery >> status: 0 handle: 2
D NimBLERemoteService: Characteristic Discovery >> status: 0 handle: 4
D NimBLERemoteService: Characteristic Discovery >> status: 14 handle: -1
D NimBLERemoteService: << Characteristic Discovery
D NimBLERemoteService: << retrieveCharacteristics()
D NimBLERemoteCharacteristic: >> retrieveDescriptors() for characteristic: 0x2a00
I (2942) NimBLE: GATT procedure initiated: discover all descriptors; 
I (2949) NimBLE: chr_val_handle=3 end_handle=5

Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x400d454a  PS      : 0x00060430  A0      : 0x800d3242  A1      : 0x3ffc6b70  
A2      : 0x3ffc9f20  A3      : 0x00000000  A4      : 0x00000000  A5      : 0x00000003  
A6      : 0x3ffc6ba0  A7      : 0x0000000c  A8      : 0x800d452c  A9      : 0x3ffc6b40  
A10     : 0x00000000  A11     : 0x7669bc1f  A12     : 0x00000000  A13     : 0x00000000  
A14     : 0x00000000  A15     : 0x3ffc2370  SAR     : 0x00000010  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x00000008  LBEG    : 0x4000c2e0  LEND    : 0x4000c2f6  LCOUNT  : 0xffffffff  


Backtrace: 0x400d4547:0x3ffc6b70 0x400d323f:0x3ffc6bc0 0x400d1a59:0x3ffc6bf0 0x400f4aa3:0x3ffc6c20 0x40096ed2:0x3ffc6c50
```

[...]

PR #885, opened 2025-02-04
fix(Client): Provide filter argument in discoverAttributes

thekurtovic:
Resolves https://github.com/h2zero/NimBLE-Arduino/issues/884


h2zero:
Thanks! Curious as to the need for a full revert instead of just providing a filter with a NULL UUID?

thekurtovic:
taken from an existing PR. maybe better if you handle it

Issue #933, opened 2025-04-08
configMAX_PRIORITIES seems to be incorrect value for freeRTOS task priority


pwilkowski:
According to freertos docs , task should have maximum priority of configMAX_PRIORITIES - 1
https://www.freertos.org/Documentation/02-Kernel/02-Kernel-features/01-Tasks-and-co-routines/03-Task-priorities

Which seems to be bigger here
```cpp
ll_task_h = xTaskCreateStatic(nimble_port_ll_task_func, "ll", NIMBLE_LL_STACK_SIZE,
                                  NULL, configMAX_PRIORITIES, ll_xStack, &ll_xTaskBuffer);
```
https://github.com/h2zero/NimBLE-Arduino/blob/62180ab2bc9c9a8b41f9fb800ea7256e89628a3a/src/nimble/porting/npl/freertos/src/nimble_port_freertos.c#L114

This issue may be related: https://github.com/h2zero/NimBLE-Arduino/pull/584

h2zero:
Thanks! not sure how it was missed and also apparently working for so long...

pwilkowski:
after enabling debugging for tasks in free rots i see this:

![Image](https://github.com/user-attachments/assets/745c7b9b-b464-4362-a056-05c17af5eff3)

i came to conclusions
1. let Ble tasks to have configurable priorities
2. by default imo it should be `configMAX_PRIORITIES -1` and `configMAX_PRIORITIES - 2` so they dont collide with built in Tmr task

h2zero:
Closing this as fixed in #891

PR #934, opened 2025-04-08
Fixing BLE priorities and making them configurable

pwilkowski:
as per https://github.com/h2zero/NimBLE-Arduino/issues/933

i checked it in my local project and it seem to work file for nrf52840

### DIFF
--- a/src/nimble/porting/npl/freertos/src/nimble_port_freertos.c
+++ b/src/nimble/porting/npl/freertos/src/nimble_port_freertos.c
@@ -41,6 +41,17 @@ static StackType_t hs_xStack[ NIMBLE_HS_STACK_SIZE ];
 static StaticTask_t hs_xTaskBuffer;
 #endif
 
+// configMAX_PRIORITIES - 1 is Tmr builtin task
+#ifndef NIMBLE_LL_PRIORITY
+#define NIMBLE_LL_PRIORITY (configMAX_PRIORITIES - 2)
+#endif
+#ifndef NIMBLE_BLE_PRIORITY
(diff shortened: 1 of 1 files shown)


h2zero:
Thanks @pwilkowski, I will merge this shortly, just need to test a few things first.

h2zero:
@pwilkowski Looks like there are some issues with this, causes errors on some of my boards. What needs to be done is the timer task needs to be lower priority than the controller task. I have added this to the nimble core 1.6 PR #891 and you can see the implementation here: https://github.com/h2zero/NimBLE-Arduino/blob/nimble-core-1.6/src/nimble/porting/npl/freertos/src/nimble_port_freertos.c

pwilkowski:
@h2zero that is weird, because it totally fixed my issues. And my timer still has highest priority. Which board are you talking about? 

Generally if possible, please make all priorities configurable. My task stack is pretty complex and i needed to fiddle around to make USB, BLE and internal tasks working fine. Solution was pretty straightforward tho, just assign each task its own unique priority and sort them out.

h2zero:
The issue was in nimble 1.6 with an nrf52833. 

I will be making all of these configurable. For now please update this PR to match the code in the 1.6 PR so I can merge this PR.

pwilkowski:
@h2zero i think i updated it correctly, please check

h2zero:
Thanks @pwilkowski, unfortunately the build failed. I have merged the nimble 1.6 update that has the changes in it though so this will no longer apply.

pwilkowski:
That is weird but I see its configurable now so all good.

PR #948, opened 2025-04-27
Reduced iram usage

doudar:
The newest versions of Arduino Core seem to treat iram differently (and use a lot more.)

This fixes iram compile issues when using pioarduino with both WiFi and NimBLE. 

Currently not well tested, but client, server, scanning and notifications, seem to work for me.

### DIFF
--- a/src/nimble/esp_port/port/include/esp_nimble_cfg.h
+++ b/src/nimble/esp_port/port/include/esp_nimble_cfg.h
@@ -18,11 +18,6 @@
 #define MYNEWT_VAL(x)                           MYNEWT_VAL_ ## x
 #define MYNEWT_VAL_CHOICE(_name, _val)          MYNEWT_VAL_ ## _name ## __ ## _val
 
-#ifndef IRAM_ATTR_64MCPU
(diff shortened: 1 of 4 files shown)


src/nimble/esp_port/port/include/esp_nimble_cfg.h
```diff
@@ -19,7 +19,7 @@
 #define MYNEWT_VAL_CHOICE(_name, _val)          MYNEWT_VAL_ ## _name ## __ ## _val
 
 #ifndef IRAM_ATTR_64MCPU
-#define IRAM_ATTR_64MCPU IRAM_ATTR
+#define IRAM_ATTR_64MCPU
```
h2zero: Lets just delete this completely, it's not used at all.

doudar: Deleted. Compiled without errors. Client, Server, Scanning and Notifications are working.

doudar:
Testing is complete from me. It's working great!!

h2zero (commented):
LGTM

PR #969, opened 2025-05-26
add XIAO

Jasionf:
image
image
image

Hello! We have successfully implemented the NImBLE routine using the XIAO ESP32 series development board. In order to better support and share this achievement, we would like to add relevant information about our development board in the README file.

Thank you very much for considering our request! Looking forward to your reply.


h2zero:
Thank you for the PR. It is not necessary to specify individual boards, instead just the chip as this library universally supports the chips that it supports regardless of the board design.

Issue #195, opened 2021-02-20
iPhone doesn't detect ESP32 as a beacon


ReyWamba:
First of all, I want to congratulate you for your work. The amount of memory that is freed is incredible.

But my iPhone doesn't detect my ESP32 as a beacon with your library.

This code works for me with the original ESP32 library, but not with yours:

<void initBLEbeacon() {
  BLEAdvertising* pAdvertising;
  pAdvertising = pServer->getAdvertising();
  pAdvertising->stop();
  // iBeacon
  BLEBeacon myBeacon;
  myBeacon.setManufacturerId(0x4c00);
  myBeacon.setMajor(deviceInfo.major);
  myBeacon.setMinor(deviceInfo.minor);
  myBeacon.setSignalPower(0xc5);
  myBeacon.setProximityUUID(BLEUUID(BEACON_UUID_REV));

  BLEAdvertisementData advertisementData;
  advertisementData.setFlags(0x06);
  advertisementData.setManufacturerData(myBeacon.getData());

  pAdvertising->setAdvertisementData(advertisementData);

  pAdvertising->start();

  esp_bt_controller_mem_release(ESP_BT_MODE_CLASSIC_BT);
  Serial.println("Inicializado beacons");
}>

Is there something i am doing wrong?

Thanks.

h2zero:
Hi @ReyWamba, thank you. I don't see anything wrong in the code. Would you be able to share a debug log?

ReyWamba:
Hello @h2zero:

I fixed the error. The problem is that the original BLEUUID () function from the ESP32 library returned the UUID string in reverse order, while yours returns it in the correct order.

That is why the iPhone was not detecting the beacon (it must "know" what UUID to look for).

Thank you very much for your speed in answering, and above all, thank you very much for your great work.

h2zero:
Sorry for the confusion with that, the NimBLE core handles that data with opposite endian from bluedroid. 

Glad you found the issue!

ReyWamba:
Hello again:

In my project, I use the ESP32 simultaneously as a BLE device with a service and as a Beacon. With the original ESP32 library, this was possible (as long as you start the Beacon first and then the service).

However, with your library, activating the Beacon disappears the service. The Beacon is detected and the BLE device is still visible but with zero services. And this happens by starting the service first and then the Beacon or vice versa.

I attach the code for both initialization routines:

`void initBLEservice() {
  BLEAdvertising* pAdvertising;
[...]

PR #971, opened 2025-05-27
Add iBeacon example

### LINKED ISSUE #911: iBeacon example version > 1.5
Hello,

Im trying to find a iBeacon example that works with the 2.2.3 version available in Arduino IDE.

I can find an example on https://github.com/h2zero/NimBLE-Arduino/blob/release/1.5/examples/Refactored_original_examples/BLE_iBeacon/BLE_iBeacon.ino but it dosn't compile with that version.

Do you have any example?

thank you,

lazd:
This example advertises an iBeacon that is recognized by Apple's CLLocationManager. I've tested this approach with a Swift app I'm building and it works great.

Based loosely on a previous [Bluedroid implementation here](https://github.com/nkolban/ESP32_BLE_Arduino/blob/master/examples/BLE_iBeacon/BLE_iBeacon.ino), modified for NimBLE.

This is the example closes #911

cc @jdbravo

### DIFF
--- a/examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
+++ b/examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
@@ -22,10 +22,9 @@ void setup() {
 
 	// Create beacon object
 	NimBLEBeacon beacon;
-	beacon.setManufacturerId(0x4C00); // fake Apple 0x004C LSB (ENDIAN_CHANGE_U16!)
 	beacon.setMajor(1);
 	beacon.setMinor(1);
-	beacon.setSignalPower(0xC5); // Not required
+	beacon.setSignalPower(0xC5); // Optional
 	beacon.setProximityUUID(BLEUUID(iBeaconUUID)); // Unlike Bluedroid, you do not need to reverse endianness here
 
 	// Extract beacon data
--- a/examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
+++ b/examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
@@ -1,0 +1,46 @@
+/**
+ *  iBeacon example
+ *
+ *  This example demonstrates how to publish an Apple-compatible iBeacon
+ *
+ *  Created: on May 26 2025
+ *      Author: lazd
+ */
+
+#include <Arduino.h>
+#include <NimBLEDevice.h>
+#include <NimBLEBeacon.h>
+
+// According to Apple, it's important to have a 100ms advertising time
+#define BEACON_ADVERTISING_TIME	160   // 100ms
+
+// Hey, you! Replace this with your own unique UUID with something like https://www.uuidgenerator.net/
+const char* iBeaconUUID = "26D0814C-F81C-4B2D-AC57-032E2AFF8642";
+
+void setup() {
+	NimBLEDevice::init("NimBLEiBeacon");
+
+	// Create beacon object
+	NimBLEBeacon beacon;
+	beacon.setManufacturerId(0x4C00); // fake Apple 0x004C LSB (ENDIAN_CHANGE_U16!)
+	beacon.setMajor(1);
+	beacon.setMinor(1);
+	beacon.setSignalPower(0xC5); // Not required
+	beacon.setProximityUUID(BLEUUID(iBeaconUUID)); // Unlike Bluedroid, you do not need to reverse endianness here
+
+	// Extract beacon data
+	NimBLEBeacon::BeaconData beaconData = beacon.getData();
+
+	// Create advertisement data
+ 	NimBLEAdvertisementData beaconAdvertisementData;
+	beaconAdvertisementData.setFlags(0x04); // BR_EDR_NOT_SUPPORTED
+	beaconAdvertisementData.setManufacturerData(reinterpret_cast<const uint8_t*>(&beaconData), sizeof(NimBLEBeacon::BeaconData));
+
+	// Start advertising
+	NimBLEAdvertising *advertising = NimBLEDevice::getAdvertising();
+	advertising->setAdvertisingInterval(BEACON_ADVERTISING_TIME);
+	advertising->setAdvertisementData(beaconAdvertisementData);
+	advertising->start();
+}
+
+void loop() {}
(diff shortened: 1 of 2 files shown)


examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
```diff
@@ -0,0 +1,46 @@
...
+	beacon.setSignalPower(0xC5); // Not required
+	beacon.setProximityUUID(BLEUUID(iBeaconUUID)); // Unlike Bluedroid, you do not need to reverse endianness here
+
+	// Extract beacon data
+	NimBLEBeacon::BeaconData beaconData = beacon.getData();
+
+	// Create advertisement data
+ 	NimBLEAdvertisementData beaconAdvertisementData;
+	beaconAdvertisementData.setFlags(0x04); // BR_EDR_NOT_SUPPORTED
+	beaconAdvertisementData.setManufacturerData(reinterpret_cast<const uint8_t*>(&beaconData), sizeof(NimBLEBeacon::BeaconData));
```
lazd: I'm curious if there's a better way to do this, I noticed I can't directly pass `beacon.getData()` as it doesn't match any of `setManufacturerData`'s signatures.

examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
```diff
@@ -0,0 +1,46 @@
...
+
+// Hey, you! Replace this with your own unique UUID with something like https://www.uuidgenerator.net/
+const char* iBeaconUUID = "26D0814C-F81C-4B2D-AC57-032E2AFF8642";
+
+void setup() {
+	NimBLEDevice::init("NimBLEiBeacon");
+
+	// Create beacon object
+	NimBLEBeacon beacon;
+	beacon.setManufacturerId(0x4C00); // fake Apple 0x004C LSB (ENDIAN_CHANGE_U16!)
```
lazd: The fact that you have to reverse endianness here can be confusing, but it's the same as Bluedroid.

examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
```diff
@@ -0,0 +1,46 @@
...
+	beacon.setMajor(1);
+	beacon.setMinor(1);
+	beacon.setSignalPower(0xC5); // Not required
+	beacon.setProximityUUID(BLEUUID(iBeaconUUID)); // Unlike Bluedroid, you do not need to reverse endianness here
```
lazd, line 28: As noted in #195, you do not need to reverse the endianness of your UUID, unlike Bluedroid.

examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
h2zero: You don't even need to set this, the default construction will do it for you 😄 .

examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
h2zero: Feature, not a bug 😄

examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
h2zero: What I see here is that this isn't needed. Instead what should be done is the beacon class should be fixed to include the flags since that is all that is happening here. They do not change in this repo as BT classic is not supported. Instead lets add this as a fixed data to the beacon data in the beacon class.

h2zero (changes requested):
Thanks, looks good, a couple of comments to work on.

examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
lazd: Sounds right to me, so what would the API end up looking like? Something like this would be nice:

```
advertising->setAdvertisementData(beacon.getAdvertisingData());
```

h2zero: Sorry for the delay getting back to you. We can change this line to simply `beaconAdvertisementData.setManufacturerData(beacon.getData());`

In order to make that work you will need to add an operator to the beacondata class so it looks like :
```
    struct BeaconData {
        uint16_t manufacturerId{0x4c00};
        uint8_t  subType{0x02};
        uint8_t  subTypeLength{0x15};
        uint8_t  proximityUUID[16]{};
        uint16_t major{};
        uint16_t minor{};
        int8_t   signalPower{};
        operator std::vector<uint8_t> () const {
            return std::vector<uint8_t>(reinterpret_cast<const uint8_t*>(this),
                                        reinterpret_cast<const uint8_t*>(this) + sizeof(BeaconData));
        }
    } __attribute__((packed));
```

lazd: Implemented! Now all that is left would be to remove the need for this

```
beaconAdvertisementData.setFlags(0x04); // BR_EDR_NOT_SUPPORTED
```

How should we go about doing that? Can we simply omit the call, or do we need to call `setFlags()` somewhere, like in the `NimBLEAdvertisementData` constructor?

examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
```diff
@@ -0,0 +1,42 @@
...
+	// Create beacon object
+	NimBLEBeacon beacon;
+	beacon.setMajor(1);
+	beacon.setMinor(1);
+	beacon.setSignalPower(0xC5); // Optional
+	beacon.setProximityUUID(BLEUUID(iBeaconUUID)); // Unlike Bluedroid, you do not need to reverse endianness here
+
+	// Create advertisement data
+ 	NimBLEAdvertisementData beaconAdvertisementData;
+	beaconAdvertisementData.setFlags(0x04); // BR_EDR_NOT_SUPPORTED
```
lazd, line 32: How do we want to handle this?

examples/NimBLE_iBeacon/NimBLE_iBeacon.ino
h2zero: I think we should leave this for now, it would overly complicate things to accommodate.

h2zero (approved):

h2zero:
Thanks!

lazd:
Thanks for the guidance on landing this pR, glad I could contribute something back!

PR #978, opened 2025-06-11
Fix build with esp32c2 and upcoming arduino core version

h2zero:
@Jason2866


Jason2866:
You nailed it. All MCUs do now compile. https://github.com/pioarduino/platform-espressif32/actions/runs/15580800440/job/43875374238

PR #1007, opened 2025-07-31
[Bugfix] OnConnectfail not called when connection not established.

### LINKED ISSUE #1006: When async-client fails to connect, the client callback `onConnectFail` doesn't get called
NimBLE-Arduino: 2.3.2
Platform:  [pioarduino](https://github.com/pioarduino/platform-espressif32) 54.03.20
Framework: Arduino 3.2.0
ESP-IDF: v5.4.1
Board: ESP32-C6
BLE Device: Xbox Wireless Controller (firmware 5.x)

On about ~10-20% connection attempts, my BLE gamepad doesn't seem to connect successfully with the board. It wouldn't be a problem if I my `onConnectFail` callback would get called, but it doesn't. 

This is the log from unsuccessfull connect attempt:

```
I NimBLEScan: New advertiser: ec:83:50:6e:5c:71
D NimBLEScanCallbacks: Discovered: Name: , Address: ec:83:50:6e:5c:71, appearance: 964, manufacturer data: 060000, serviceUUID: 0x1812, txPower: 20
I NimBLEScan: Scan response from: ec:83:50:6e:5c:71
DEBUG Device discovered, address: ec:83:50:6e:5c:71, address type: 0, name: Xbox Wireless Controller
INFO Attempting to connect to a device, address: ec:83:50:6e:5c:71
D NimBLEClient: >> connect(ec:83:50:6e:5c:71)
D NimBLEScan: >> stop()
D NimBLEScan: << stop()
D NimBLEScan: erase device: ec:83:50:6e:5c:71
D NimBLEClient: >> handleGapEvent BLE_GAP_EVENT_DISCONNECT
D NimBLEClient: my inserted log: disconnect reason is 0x023e
D NimBLEClient: my inserted log: event->disconnect.conn.conn_handle is 0

no more logs
````

I was digging into the library code so I've inserted two log lines into the `NimBLEClient` (inserted just after the line 935). The logged reason is `Connection Failed to be Established` according to [docs](https://mynewt.apache.org/latest/network/ble_hs/ble_hs_return_codes.html). While the `conn_handle` from event object is 0.

https://github.com/h2zero/NimBLE-Arduino/blob/20097ffab9379183b07b9c301799521b0697d5d0/src/NimBLEClient.cpp#L934-L940

What happens next is that `NimBLEDevice::getClientByHandle` returns `nullptr`, so the function `NimBLE::handleGapEvent` simply returns `0` without notyfing the applicaiton code about an unsuccessful connection attempt.

My application code fragments:
* client [`connect` implementation](https://github.com/tbekas/BLE-Gamepad-Client/blob/d790821001bc49ba145f7497f8d4141108ec0979/src/BLEScanCallbacksImpl.cpp#L71)
* client callbacks [`onConnectFail` implementation](https://github.com/tbekas/BLE-Gamepad-Client/blob/1847b744dbd2ba216f974455bde331c304c5de33/src/BLEClientCallbacksImpl.cpp#L14)

For context - this is how logs look like on a successful connection attempt (80-90% of the time).

```
I NimBLEScan: New advertiser: ec:83:50:6e:5c:71
D NimBLEScanCallbacks: Discovered: Name: , Address: ec:83:50:6e:5c:71, appearance: 964, manufacturer data: 060000, serviceUUID: 0x1812, txPower: 20
I NimBLEScan: Scan response from: ec:83:50:6e:5c:71
DEBUG Device discovered, address: ec:83:50:6e:5c:71, address type: 0, name: Xbox Wireless Controller
INFO Attempting to connect to a device, address: ec:83:50:6e:5c:71
D NimBLEClient: >> connect(ec:83:50:6e:5c:71)
D NimBLEScan: >> stop()
D NimBLEScan: << stop()
D NimBLEScan: erase device: ec:83:50:6e:5c:71
D NimBLEClient: >> handleGapEvent BLE_GAP_EVENT_CONNECT
DEBUG Connected to a device, address: ec:83:50:6e:5c:71
D NimBLEClient: >> secureConnection()
D NimBLEClient: >> handleGapEvent BLE_GAP_EVENT_LINK_ESTAB
D NimBLEClient: >> handleGapEvent BLE_GAP_EVENT_DATA_LEN_CHG
D NimBLEClient: >> handleGapEvent BLE_GAP_EVENT_MTU
I NimBLEClient: mtu update: mtu=23
D NimBLEClientCallbacks: onMTUChange: default
D NimBLEClient: << handleGapEvent
D NimBLEClient: exchangeMTUCb: status=0, mtu=23
D NimBLEClient: >> handleGapEvent BLE_GAP_EVENT_PARING_COMPLETE
D NimBLEClient: >> handleGapEvent BLE_GAP_EVENT_ENC_CHANGE
INFO Bonded successfully with a device, address: ec:83:50:6e:5c:71
```

### PR DESCRIPTION
**Pull request by h2zero**
Workaround for when the disconnect event is sent when no connection has been established. Espressif changed this from a connect event with error code to disconnect event.

Fixes #1006

### DIFF
--- a/src/NimBLEClient.cpp
+++ b/src/NimBLEClient.cpp
@@ -933,8 +933,15 @@ int NimBLEClient::handleGapEvent(struct ble_gap_event* event, void* arg) {
 
     switch (event->type) {
         case BLE_GAP_EVENT_DISCONNECT: {
+            NimBLEConnInfo peerInfo;
+            rc = ble_gap_conn_find(event->disconnect.conn.conn_handle, &peerInfo.m_desc);
+            if (rc != 0) {
+                NIMBLE_LOGE(LOG_TAG, "Failed to find connection info; rc=%d", rc);
+                return 0;
+            }
+
             // workaround for bug in NimBLE stack where disconnect event argument is not passed correctly
-            pClient = NimBLEDevice::getClientByHandle(event->disconnect.conn.conn_handle);
+            pClient = NimBLEDevice::getClientByPeerAddress(peerInfo.m_desc.peer_id_addr);
             if (pClient == nullptr) {
                 return 0;
             }
@@ -960,7 +967,9 @@ int NimBLEClient::handleGapEvent(struct ble_gap_event* event, void* arg) {
             pClient->m_asyncSecureAttempt = 0;
 
             // Don't call the disconnect callback if we are waiting for a connection to complete and it fails
-            if (rc != (BLE_HS_ERR_HCI_BASE + BLE_ERR_CONN_ESTABLISHMENT) || pClient->m_config.asyncConnect) {
+            if (rc == (BLE_HS_ERR_HCI_BASE + BLE_ERR_CONN_ESTABLISHMENT) && pClient->m_config.asyncConnect) {
+                pClient->m_pClientCallbacks->onConnectFail(pClient, rc);
+            } else {
                 pClient->m_pClientCallbacks->onDisconnect(pClient, rc);
             }
 

### CODE REVIEW

File: src/NimBLEClient.cpp [hunk 1]
```diff
@@ -933,8 +933,15 @@ int NimBLEClient::handleGapEvent(struct ble_gap_event* event, void* arg) {
 
     switch (event->type) {
         case BLE_GAP_EVENT_DISCONNECT: {
+            NimBLEConnInfo peerInfo;
+            rc = ble_gap_conn_find(event->disconnect.conn.conn_handle, &peerInfo.m_desc);
```
tbekas: When the connection attempt fails, the first argument to this function call seems to be `0`, thus this function returns non-zero return code.

Maybe we could try finding the client using peer address from `event->disconnect.conn.peer_id_addr` instead?

h2zero: Thanks, it should be resolved now.

tbekas: I tested it and it works flawlessly. Thank you!

Issue #918, opened 2025-03-21
ERROR "<< sendValue: failed to allocate mbuf" some time after the start of sending notifications. v2.2.3


truefalsenone:
Hello,

I am using the ESP32-C3. My project has a BLE-client and a BLE-server. The NimBLE-Arduino version is v2.2.3.

The BLE-server sends notifications with a payload of 40 bytes 2 times per second. However, after some time (about 50 minutes), the notifications stop being sent, and in the server log I see an error: "<< sendValue: failed to allocate mbuf." 

```
NimBLECharacteristic.cpp

// Must re-create the data buffer on each iteration because it is freed by the calls bellow.
os_mbuf* om = ble_hs_mbuf_from_flat(value, length);
if (!om) {
    NIMBLE_LOGE(LOG_TAG, "<< sendValue: failed to allocate mbuf");
    return false;
}
```

If I reconnect, the notifications are successfully sent again. 
70 KB are free in a heap all the time.

h2zero:
You should wait until the `onStatus` callback is invoked (checking for success of the last notification, rc==0) before sending the next notification. The issue in this case is that there is a connection issue between the devices and each time a notification is sent it consumes a buffer but if that doesn't get sent then the buffer doesn't get freed.

h2zero:
Closing this as complete.

PR #1052, opened 2025-10-23
Refactor notify/indicate

h2zero:
This refactors the handling of sending notifications and indications for greater efficiency.
* Adds client subscription state tracking to NimBLECharacteristic rather than relying on the stack.
* Notifications/indications are now sent directly, no longer calling the callback to read the values. This avoids delays and flash writes in the stack, allowing for greater throughput.
* Notifications will now wait up to 10ms if buffers are full to reattempt transmission before failing. 
* Adds a new overloaded callback to NimBLECharacteristicCallbacks for the notification/indication onStatus method that provides a NimBLEConnInfo reference.

#918

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,7 +61,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.x"
+          python-version: "3.13"
       - name: Install platformio
         run: |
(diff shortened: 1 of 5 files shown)


barosl:
@h2zero Thanks for the refactor, this also fixed the issue where clients that are pending authentication pairing could subscribe to and receive notifications.

PR #1055, opened 2025-10-24
Fix incorrect byte size calculation in NimBLEAttValue::setValue for container types

srgg:
# Correct byte size handling for container types in NimBLEAttValue::setValue

## Summary
Improves `NimBLEAttValue::setValue` so that all container types correctly calculate byte size before updating an ATT value. The fix fully addresses incorrect buffer sizing when the container element type is larger than one byte.

## Problem
The previous implementation used `v.size()` as the byte length. This works only when elements are 1 byte. For multi-byte element types (for example `uint16_t`), this caused:

- Incorrect memory allocation
- Data truncation or potential buffer overflow
- Wrong BLE payload size reported to the stack

Example:  
`std::vector<uint16_t>` with 10 elements  
- Before: 10 bytes (incorrect)  
- Now: 20 bytes (correct: 10 × 2 bytes)

## Changes Made

1. **Added `Has_value_type` trait** (lines 76–82)
   - C++11-compatible SFINAE trait that detects presence of `value_type`
   - Works with all standard containers (`std::vector`, `std::array`, `std::string`, etc.)

2. **Updated pre-C++17 templates** (lines 282–314)
   - Split into two overloads:
     - With `value_type`: size in bytes = `size() × sizeof(value_type)`
     - Without `value_type`: use `size()` directly (assumed to already represent bytes)

3. **Updated C++17+ template** (lines 324–329)
   - Uses nested `if constexpr` to branch at compile time
   - Same logic as pre-C++17 implementation but cleaner and safer

## Behavior Examples

| Type                         | size() | Byte calculation | Result |
|-----------------------------|:------:|:----------------:|:------:|
| `std::vector<uint8_t>`      | 10     | 10×1             | 10 ✓   |
| `std::vector<uint16_t>`     | 10     | 10×2             | 20 ✓   |
| `std::vector<uint32_t>`     | 10     | 10×4             | 40 ✓   |
| `std::string`               | chars  | `Has_c_str_length` override | ✓ |
| Custom container without `value_type` | N elements | Uses size() directly | ✓ |

## Impact
- Correct, safe sizing for all container types
- Avoids data corruption and memory errors
- No API changes required
- Backwards compatible behavior for byte-sized types

### DIFF
--- a/src/NimBLECharacteristic.h
+++ b/src/NimBLECharacteristic.h
@@ -112,7 +112,27 @@ class NimBLECharacteristic : public NimBLELocalValueAttribute {
     }
 
     /**
-     * @brief Template to send a notification with a value from a class that has a data() and size() method.
+         * @brief Template to send a notification with a value from a class that has a data() and size() method with value_type.
+     * @param [in] v The value to send.
+     * @param [in] connHandle Optional, a connection handle to send the notification to.
+     * @details Correctly calculates byte size for containers with multi-byte element types.
+     */
+    template <typename T>
+#  ifdef _DOXYGEN_
+    bool
+#  else
+    typename std::enable_if<Has_data_size<T>::value && Has_value_type<T>::value, bool>::type
+#  endif
+    notify(const T& v, uint16_t connHandle = BLE_HS_CONN_HANDLE_NONE) const {
+        return notify(
+            reinterpret_cast<const uint8_t*>(v.data()),
+            v.size() * sizeof(typename T::value_type),
+            connHandle
+        );
+    }
+
+    /**
+     * @brief Template to send a notification with a value from a class that has a data() and size() method without value_type.
      * @param [in] v The value to send.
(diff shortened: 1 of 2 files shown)


h2zero:
Thanks! I hadn't thought of this use case before. Could you also apply this to the `NimBLERemoteAttribute::writeValue` and `NimBLECharacteristic::notify`/`NimBLECharateristic::indicate` templates as well?

srgg:
e80a7fc314f49dce1e2857d612fb60fff104eefa extends the previous fix for `NimBLEAttValue::setValue` to additional template methods across the NimBLE library. These changes ensure that containers with multi-byte element types are properly sized when written, notified, or indicated.

srgg:
Hi @h2zero, have you had a chance to review the recent changes? Are they good to go, or do you see anything that needs adjustment?

h2zero:
Sorry, been busy. Changes look good, I will merge soon.

PR #1070, opened 2025-11-24
Added Stream Class Examples. Fixes a couple bugs in Stream Class. 

doudar:
Creates examples demonstrating the new NimBLEStream classes that provide Arduino Stream interface (print(), println(), read(), available()) over BLE characteristics.
Examples Added

    NimBLE_Stream_Echo - Minimal echo server (52 lines), demonstrates basic API
    NimBLE_Stream_Server - Full server implementation with connection management, MTU negotiation, bidirectional communication
    NimBLE_Stream_Client - Client with automatic discovery, reconnection, pairs with server example

Usage

NimBLEStreamServer bleStream;

void setup() {
    NimBLEDevice::init("BLE-Device");
    NimBLEDevice::createServer();
    
    // Initialize with custom UUIDs, enable writes
    bleStream.init(serviceUUID, charUUID, true, false);
    bleStream.begin();
    
    NimBLEDevice::getAdvertising()->start();
}

void loop() {
    if (bleStream.hasSubscriber()) {
        bleStream.println("Hello via BLE");  // Just like Serial
        
        if (bleStream.available()) {
            char c = bleStream.read();
            bleStream.write(c);  // Echo back
        }
    }
}

Documentation

    Individual README per example with usage instructions
    STREAM_EXAMPLES.md overview covering API, use cases, troubleshooting

Source Fixes

Fixed issues in merged NimBLEStream implementation:

    Missing <cstring> include when Arduino unavailable
    Incorrect return value in uart_log_printfv (returned calculated length instead of written bytes)
    Format specifier mismatch (%u → %zu for size_t)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 docs/doxydocs
 .development
+_codeql_detected_source_root
--- a/examples/NimBLE_Stream_Client/NimBLE_Stream_Client.ino
+++ b/examples/NimBLE_Stream_Client/NimBLE_Stream_Client.ino
@@ -1,0 +1,215 @@
+/**
+ *  NimBLE_Stream_Client Example:
+ *
+ *  Demonstrates using NimBLEStreamClient to connect to a BLE GATT server
+ *  and communicate using the Arduino Stream interface.
+ *  
+ *  This allows you to use familiar methods like print(), println(), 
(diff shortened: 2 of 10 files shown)


h2zero:
Nice, thanks!

h2zero:
Closing this as I have squashed/cherry-picked this into the stream-class PR now, thanks!

PR #1082, opened 2026-02-04
[Bugfix] Peripheral connection event delay/missing disconnect event

h2zero:
This removes some incorrect upstream changes that would allow a peripheral/server device to receive events from connected client requests without the application being aware of the connection.

The code was changed from the original mynewt event flow in that the connection event is not sent until a response from a request to read the remote version/features where originally the event was sent before these responses were received.

This reverts this situation for the peripheral role because when the connection is made in this case it is fully established and there is no need to wait for the responses to the above requests. In addition the delay caused by these was allowing connected peers to interact without the application knowing of the connection, creating unexpected issues.

One of those issues is the client could disconnect before the peripheral receives a response to the above requests and the changes that were made would prevent the disconnect event from being sent because the connection was not considered established.

Finally, the connection reattempt code was checking for a connection establishment failure when the device was acting as a slave, which is incorrect as that condition cannot happen if the spec is followed. The controller should never send this event to a peripheral device waiting for connection, it should just stay advertising.

This patch removes all of the unnecessary code that was added and reverts the flow to the original mynewt flow for peripheral devices.

### DIFF
--- a/src/nimble/nimble/host/src/ble_gap.c
+++ b/src/nimble/nimble/host/src/ble_gap.c
@@ -296,7 +296,6 @@ static uint8_t pawr_adv_handle;
 static uint16_t pawr_sync_handle;
 #endif
 
-int slave_conn[MYNEWT_VAL(BLE_MAX_CONNECTIONS) + 1];
 static void ble_gap_update_entry_free(struct ble_gap_update_entry *entry);
 
 #if NIMBLE_BLE_CONNECT
@@ -1557,8 +1556,6 @@ ble_gap_conn_broken(uint16_t conn_handle, int reason)
     struct ble_gap_update_entry *entry;
     struct ble_gap_snapshot snap;
     struct ble_gap_event event;
-    struct ble_hs_conn *conn;
-    bool send = 1;
     int rc;
 
     memset(&event, 0, sizeof event);
@@ -1595,19 +1592,6 @@ ble_gap_conn_broken(uint16_t conn_handle, int reason)
 #endif
     ble_hs_flow_connection_broken(conn_handle);;
 
-    ble_hs_lock();
-    conn = ble_hs_conn_find(conn_handle);
-    ble_hs_unlock();
-
-    // Send disconnect event in slave role if connect was sent
-    if ((conn != NULL) &&  !(conn->bhc_flags & BLE_HS_CONN_F_MASTER)) {
-        if (slave_conn[conn_handle]) {
-            slave_conn[conn_handle] = 0;
-	} else {
-	    send = 0;
-	}
-    }
-
     ble_hs_atomic_conn_delete(conn_handle);
 
     g_max_tx_time[conn_handle] = 0;
@@ -1618,10 +1602,8 @@ ble_gap_conn_broken(uint16_t conn_handle, int reason)
     event.type = BLE_GAP_EVENT_DISCONNECT;
     event.disconnect.reason = reason;
 
-    if (send) {
-        ble_gap_event_listener_call(&event);
-        ble_gap_call_event_cb(&event, snap.cb, snap.cb_arg);
-    }
+    ble_gap_event_listener_call(&event);
+    ble_gap_call_event_cb(&event, snap.cb, snap.cb_arg);
 
     STATS_INC(ble_gap_stats, disconnect);
 #endif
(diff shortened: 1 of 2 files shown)


doudar:
### Ahoy Matey! Here be me feedback fer PR #1082 like ye requested!

---

### 1. **Summary o' Changes**:
This here pull request aims to fix a bug fer the nimble-arduino library by reworking the peripheral server connection flow and disconnect event handling. The swabbin' that the code received removes upstream changes that delayed the connection event fer a peripheral/server device until a remote version/features request response was received from the client. Instead, the event shan't wait fer such responses anymore, as the original `mynewt` flow used to send events prior to those responses.

Key changes:  
- **Improved event flow**: It simplifies and refines the peripheral role's handling o' connection and disconnection events. 
- Removed old variables (`slave_conn`) and logic, declutterin' unnecessary checks.
- Fixed roles handling (Master/Slave) to ensure events are handled more appropriately, restorin’ original functionalities fer both connection and disconnection events.
- Bugfix fer sending disconnect events properly, removin’ outdated constraints on "slave_conn".
- Cleaned up callbacks and redundant code sprinklin’.

---

### 2. **Code Quality:**
#### Good:
- **Back to the basics**: Ye reverted misguided changes and restored the natural order, makin’ things a lot cleaner in the code’s flow.
- Ye removed that pesky `slave_conn` global state, which limits the scope o’ accidental misuse an’ reduces memory footprint.
- A clear reduction in redundant conditionals an’ unnecessary complexity is here — good work!
- Following modular practices fer event callbacks (`ble_gap_event_listener_call`/`ble_gap_call_event_cb`) makes it easier fer buccaneers down the line to handle events.

#### Concerns:
- **Lock ’n’ unlock patterns**: All calls to `ble_hs_lock`/`ble_hs_unlock` seem proper, aye, but be sure they’re watertight to avoid any concurrency flukes. Mutiny be risin’ quick if someone forgets to lock properly elsewhere, mark me words. 
- **Comments clarity**: The code yer writin’ has decreased its number of detailed comments ever so slightly. For coders who set sail into unfamiliar waters, be sure to leave yer compass marks. Ye see, this makes life easier fer swashbucklers new onboard!
- **Magic numbers**: The code considers statuses like `BLE_ERR_CONN_ESTABLISHMENT` and `BLE_ERR_CONN_SPVN_TMO` explicitly, aye. But perhaps adding macros or constants to *semanticize* these error codes further might be worth doin’. 'Twill help scalawags later understand what those values mean without divin’ to the documentation!

---

### 3. **Security ’n’ Performance Notes:**
#### Security:
- By eliminatin’ global state (like `slave_conn`), ye minimized thy risk o’ unexpected behavior from external alterations. Good shot, matey!
- Any risks with these changes now be likely limited to rogue or unchecked application-specific client behavior.
- Be mindful o’ the fixed flags used fer Master/Slave (e.g., `BLE_HS_CONN_F_MASTER`). Hardcoded flag manipulations might lead to fragile assumptions if there be upstream changes to the flag definitions.

#### Performance:
- Referrin' and checkin' `conn_handle` still carries natural overhead, but ye seem to have trimmed most o’ the excess code and logic, which saves cycles fer the ship’s engine. Bravo!
- Consider performant event dispatch optimization: If the callbacks `ble_gap_event_listener_call` and `ble_gap_call_event_cb` be insufficient fer heavy traffic *when scaling*, a more advanced publish-subscribe model might steer the ship better.

---

### 4. **Suggestions fer Improvement**:
1. Enhancify yer comments fer clarity. They be sparse like treasure in the uncharted seas. Explain key conditions fer Master/Slave behavior, they can be tricky fer new deckhands.
2. Consider grouping or externalizing common flag definitions and error statuses into one single header file, reducing room fer error in repetitious magic numbers.
3. Another tidy-up suggestion: Avoid redundant semicolons (`;;;;`). It's common crew courtesy to steer clear, fer too much flourishes can confuse the ship's compiler. Ye left one stray line here in `ble_gap_conn_broken`.

---

### 5. **Overall Assessment**:
Ye changes restore sanity to event handling, removin’ unnecessary waitin’ fer remote responses and bringing clarity to the flow. The refactor simplifies the code, improves maintainability, and patches unintended behaviors like missed disconnect events fer peripheral roles. However, the crew might need a better map in the form o' comments fer smooth sailin’ in future voyages through this code. With slight improvements, this PR stands sturdy and seaworthy.

h2zero:
Closing this in favor of #1104

Issue #1164, opened 2026-06-28
Task watchdog boot-loop when debug log messages enabled


vlucent:
## Config

on platformio, tested with pioarduino: ```53.03.13 (arduino core 3.3.9)``` and ```55.03.39 (arduino core 3.3.9)```
board: ```esp32-s3-devkitc-1 ```

## Reproduce
app: ```examples\NimBLE_Scan_Continuous```
single change to nimconfig.h:
```
#define MYNEWT_VAL_NIMBLE_CPP_LOG_LEVEL 4
```

### pioarduino v53.03.13 (arduino core 3.3.9)
```
E (10482) task_wdt: Task watchdog got triggered. The following tasks/users did not reset the watchdog in time:
E (10482) task_wdt:  - IDLE0 (CPU 0)
E (10482) task_wdt: Tasks currently running:
E (10482) task_wdt: CPU 0: nimble_host
E (10482) task_wdt: CPU 1: loopTask
E (10482) task_wdt: Aborting.
E (10482) task_wdt: Print CPU 0 (current core) backtrace


Backtrace: 0x42003be6:0x3fcb40f0 0x4200772d:0x3fcb4170 0x4200c461:0x3fcb41e0 0x4200c2ef:0x3fcb4220 0x4200c6d3:0x3fcb4240 0x4200b5a5:0x3fcb4260 0x42011b35:0x3fcb4280 0x42002ecc:0x3fcb42a0 0x4038089a:0x3fcb42c0
  #0  0x42003be6 in NimBLEScan::handleGapEvent(ble_gap_event*, void*) at lib/NimBLE-Arduino/src/nimble/porting/npl/freertos/include/nimble/nimble_npl_os.h:642
      (inlined by) NimBLEScan::handleGapEvent(ble_gap_event*, void*) at lib/NimBLE-Arduino/src/NimBLEScan.cpp:275
  #1  0x4200772d in ble_gap_disc_report at lib/NimBLE-Arduino/src/nimble/nimble/host/src/ble_gap.c:1018
      (inlined by) ble_gap_rx_adv_report at lib/NimBLE-Arduino/src/nimble/nimble/host/src/ble_gap.c:1607
  #2  0x4200c461 in ble_hs_hci_evt_le_adv_rpt at lib/NimBLE-Arduino/src/nimble/nimble/host/src/ble_hs_hci_evt.c:564
[...]

PR #1165, opened 2026-06-29
fix infinite recursion bug for ticks to ms conversion

vlucent:
also fixes the watchdog boot loop as that's also why it was timing out #1164


h2zero:
Thanks!, not sure where that came from...

vlucent:
To be fair, those function names all look very similar. 

Also, fun fact: it's almost exactly 1 year ago since the initial commit.

PR #1167, opened 2026-06-29
NimBLELog timestamps

vlucent:
adds timestamps to NimBLELog messages :)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added timestamps to Arduino console log messages, making log output easier to trace and read.
  * Fixed a time-conversion issue in the FreeRTOS porting layer that could cause incorrect behavior during tick-to-millisecond conversion.

### DIFF
--- a/src/NimBLELog.h
+++ b/src/NimBLELog.h
@@ -147,27 +147,28 @@
 
 # else
 #  include "nimble/console/console.h"
+#  include "esp32-hal.h"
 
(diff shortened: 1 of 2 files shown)


h2zero:
Thanks but I don't think this would be a good addition as it's better handled on the logging terminal side.

vlucent:
Yeah that's a good point :(

PR #1181, opened 2026-08-23
Add sm_sec_req_auto_pair: configurable auto-pairing on Security Request

tonywestonuk:
When a peripheral we have no keys for sends an SMP Security Request, the host currently sends a Pairing Request immediately, synchronously from ble_sm_sec_req_rx(), with no way for the application to intervene.

Some peripherals (e.g. the Okida OT-2000 oven timer module, BLE 4.0 era) cannot handle a Pairing Request arriving in the same instant as their own Security Request: they never send a Pairing Response. The same devices pair on the first attempt with Bluedroid, macOS and iOS, which hand the Security Request to the application and initiate pairing a few milliseconds later.

Core Spec Vol 3, Part H, 2.4.6 leaves this to the central ("may"). This adds ble_hs_cfg.sm_sec_req_auto_pair (syscfg BLE_SM_SEC_REQ_AUTO_PAIR, default 1 = unchanged behaviour). When cleared, a Security Request from a peer we have no keys for is ignored and the application initiates pairing itself, e.g. via NimBLEClient::secureConnection(). Peers we already have keys for are handled exactly as before. Exposed as NimBLEDevice::setSecurityAutoPairOnSecReq().

Verified on ESP32-C3 / arduino-esp32 3.3.11: with the flag cleared the oven pairs first try; with stock behaviour it fails regardless of MTU exchange, connection interval, key distribution, address type or SC/legacy choice.

## Summary by CodeRabbit

* **New Features**
  * Added a setting to control whether security requests from unpaired devices automatically initiate pairing.
  * Automatic pairing remains enabled by default for compatibility.
  * Applications can disable automatic pairing and initiate pairing explicitly when needed.


tonywestonuk:
Fixes #1182

h2zero:
Thanks @tonywestonuk, this looks good but I don't like altering the nimble core files outside of the porting layer.
If you could open this PR at https://github.com/apache/mynewt-nimble and it gets merged there then I will merge this here.

tonywestonuk:
Opened upstream as apache/mynewt-nimble#2287. Hopefully they'll accept it, will post here when its merged.

PR #1185, opened 2026-09-04
host: do not deinit ble_hs_timer while its event may be queued (#1184)

### LINKED ISSUE #1184: ble_hs_timer_reset() deinits the host timer callout while its event may be queued, host task then calls a NULL ev->fn
Reporting a field crash on NimBLE-Arduino 2.5.1, ESP32-S3, arduino-esp32 3.3.x, peripheral role (advertising, no connection at the time). We have a coredump and a root cause, but no reliable reproducer.

### Symptom

`NimBLEDevice::deinit(true)` occasionally resets the chip. The crashed task is `nimble_host`:

```
Crashed task: 'nimble_host'
exccause  0x14 (InstFetchProhibitedCause)
pc        0x0
a10       -> ble_hs_timer + 8
```

`pc 0x0` with `InstFetchProhibited` is a call through a NULL function pointer. The faulting instruction is in `nimble_port_run()`:

```asm
l32i   a8, a10, 4     ; a8 = ev->fn
callx8 a8             ; a8 == 0
```

`a10` is the event that was just dequeued, and it resolves to `ble_hs_timer + 8`, i.e. the `ev` member of the host timer callout. So `ble_hs_timer.ev.fn` was NULL when the host task ran the event.

The task that called `deinit()` was blocked at the same moment in `NimBLEDevice::deinit(true)` -> `nimble_port_stop()` -> `ble_hs_stop()` -> `ble_gap_preempt()` -> `ble_gap_adv_stop_no_lock()`, waiting on the controller semaphore.

### Cause

1. `ble_hs_stop_begin()` (`src/nimble/nimble/host/src/ble_hs_stop.c`) sets `ble_hs_enabled_state = BLE_HS_ENABLED_STATE_STOPPING` and then calls `ble_hs_timer_resched()`.
2. `ble_hs_is_enabled()` returns true only for `_ON`, so it is now false.
3. `ble_hs_timer_reset()` (`src/nimble/nimble/host/src/ble_hs.c:445-447` on master) therefore takes the branch:
   ```c
   ble_npl_callout_stop(&ble_hs_timer);
   ble_npl_callout_deinit(&ble_hs_timer);
   ```
4. `npl_freertos_callout_deinit()` calls `ble_npl_event_deinit(&co->ev)` and then `memset(co, 0, sizeof(struct ble_npl_callout))`, so `ev.fn` becomes NULL.
5. If the timer had already expired, that same `ev` is **already queued on `g_eventq_dflt`**. The memset does not remove it from the queue. The host task dequeues it and calls `ev->fn`.

`ble_npl_eventq_remove()` exists in the FreeRTOS port and would cover step 4, but is not called.

### The `deinit()` call does not exist upstream

Checked 2026-09-01:

| tree | `ble_hs_timer_reset()`, not-enabled branch |
| --- | --- |
| apache/mynewt-nimble master | `ble_npl_callout_stop()` only |
| espressif/esp-nimble master | `ble_npl_callout_stop()` only |
| espressif/esp-nimble @ `f566133`, `e3cbdc0`, `70439dd` | `stop()` **and** `deinit()` |
| h2zero/NimBLE-Arduino master | `stop()` **and** `deinit()` |

The extra `deinit()` appears in older esp-nimble snapshots and espressif have since removed it, so this looks like a stale sync rather than a deliberate divergence.

Removing it should leak nothing: `ble_hs_deinit()` in the same file already ends with `ble_npl_callout_deinit(&ble_hs_timer)`, at a point where the host has stopped and the event queue is drained.

### Reproducer

We do not have one, and we want to be upfront about that. The window needs the host timer event to be in flight in the same instant the stop begins. This fired once on a device that had done hundreds of init/deinit cycles, and we found it from the coredump rather than from a test.

Happy to send a PR with the one-line removal if you would prefer that to waiting for a core re-sync.

rfordinal:
Fixes #1184. One line removed. This is the PR offered in that issue.

## The crash

Reproduced twice on real hardware, ESP32-S3, from two independent coredumps
(2026-09-01 and 2026-09-04), both while the application called
`NimBLEDevice::deinit(true)`:

```
Crashed task: 'nimble_host'
exccause  0x14 (InstFetchProhibitedCause)
excvaddr  0x0
pc        0x0
a0        0x82053472   -> NimBLEDevice::host_task+6
a10       0x3fca5878   -> ble_hs_timer + 8
```

`pc = 0` with `InstFetchProhibited` is a call through a NULL function pointer.
The faulting instruction is in `nimble_port_run()`:

```asm
0x420638da:  l32i   a8, a10, 4     ; a8 = ev->fn
0x420638dd:  callx8 a8             ; a8 == 0
```

`a10` is the dequeued event, and it resolves to `ble_hs_timer + 8`, the `ev`
member of the host's own timer callout. So `ble_hs_timer.ev.fn` was NULL when
the host task ran the event.

In the second dump `a10` holds the identical value and the code registers are
offset by a constant, so it is the same event and the same call site.

## Why it happens

1. `ble_hs_stop_begin()` (`ble_hs_stop.c`) sets
   `ble_hs_enabled_state = BLE_HS_ENABLED_STATE_STOPPING`, then calls
   `ble_hs_timer_resched()` on the next line.
2. `ble_hs_is_enabled()` returns true only for `_ON`, so it is now false.
3. `ble_hs_timer_reset()` therefore takes the disabled branch and calls
   `ble_npl_callout_stop()` **and** `ble_npl_callout_deinit()`.
4. `npl_freertos_callout_deinit()` does `ble_npl_event_deinit(&co->ev)` and then
   `memset(co, 0, sizeof(struct ble_npl_callout))`. `ev.fn` is now NULL.
5. If the timer had already expired, its `ev` is **already queued** on
   `g_eventq_dflt`. The memset does not remove it. `nimble_port_run()` dequeues
   it and calls `ev->fn`.

`ble_npl_eventq_remove()` exists in that port and is what step 4 would have to
call. Removing the `deinit()` is the smaller fix, and it matches what the other
two trees already do.

## Why removing it leaks nothing

`ble_hs_deinit()` already ends with `ble_npl_callout_deinit(&ble_hs_timer)`.
That runs after the host has stopped and the queue is drained, which is the
right time. The call in `ble_hs_timer_reset()` is redundant as well as harmful.

## Comparison with the other trees

| tree | `ble_hs_timer_reset()` when not enabled |
|---|---|
| apache/mynewt-nimble master | `ble_npl_callout_stop()` only |
| espressif/esp-nimble master | `ble_npl_callout_stop()` only |
| espressif/esp-nimble @ `f566133`, `e3cbdc0`, `70439dd` | `stop()` **and** `deinit()` |
| h2zero/NimBLE-Arduino master | `stop()` **and** `deinit()` |

The `deinit()` came from esp-nimble, espressif has since dropped it, and this
tree still carries the older snapshot.

## Testing

The removal itself is not yet run on hardware. What is verified is the fault, the
faulting instruction and the event identity, from two coredumps on ESP32-S3
(NimBLE-Arduino 2.5.1, ESP-IDF 5.5.2), plus the five-step chain read off the
vendored source. We are carrying this exact commit in a pinned fork and will
report back once it has soak-tested on our boards.

Happy to adjust the shape of the fix if you would rather remove the queued event
than skip the deinit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Bluetooth host timer handling when the host is disabled, allowing the timer to be reused when the host is re-enabled.


h2zero:
Thanks!

PR #1186, opened 2026-09-06
[Bugfix] Fix corrupted remote attribute reads with chained mbufs

lowbaud:
Fixes characteristic reads when the response is split across multiple mbufs.

I ran into this with MTU 512 while reading a larger characteristic. The data was correct up to byte 279 and corrupted after that. This likely won’t show up with smaller MTUs where the response fits into a single mbuf.

From what I can tell, the issue is that the full packet length is copied from the first mbuf's data pointer instead of walking the mbuf chain. I don't work with NimBLE internals regularly, so please double-check that assumption.

This changes the read handling so chained mbufs are copied correctly instead of being treated as one contiguous buffer.

I targeted release/2.5 since that’s where I reproduced the issue and wanted to keep the change minimal. Happy to retarget this to master if that’s preferred.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when reading remote Bluetooth attribute values split across multiple data segments.
  * Detects incomplete value assembly and reports an appropriate resource error instead of returning potentially invalid data.
  * Successfully assembled values now complete without unnecessary additional processing.


h2zero:
@coderabbitai review

h2zero:
@lowbaud Thank you for your submission.
Could you tell me what MCU you are using that prompted this? looking at the code for `os_mbuf_append` I can see it already does this internally but some MCU's use the espressif version and it may not.

lowbaud:
I'm using an ESP32-S3 with pioarduino.

You mentioned os_mbuf_append(), but the call here is actually NimBLEAttValue::append(). It's getting the length of the whole mbuf chain but only a pointer to the first buffer's data, which is why I added the loop. Am I missing something?

h2zero:
You're right @lowbaud, sorry for the confusion, I was looking at the local attribute code.

This fix looks good, I'm surprised this hasn't come up before.

